PR #9, opened 2019-12-02
fix(SearchResults): title fallback and footerbar value max length

Steven-Gassert:
This changes the result title fallback. Also this will make it so that result titles and collection names will be cut off with `...` if they begin to overflow. See this gif

hover-title

related to issue: https://github.ibm.com/Watson-Discovery/disco-widgets/issues/423 and https://github.ibm.com/Watson-Discovery/wd-issues/issues/1180

You can test this by starting storybook and changing the value of `title` to be any large value inside of `packages/discovery-react-components/src/components/SearchResults/__fixtures__/searchResults.ts`. Next open the SearchResults component and change the value of `resultTitleField` to be `title`. The title being displayed should be cutoff and `...` should be appended. You can do the same with `name` inside of `collectionsResponse.ts`

Also see the description of https://github.ibm.com/Watson-Discovery/wd-issues/issues/1180
and the new unit tests added to confirm new fallback functionality

I figured in-depth documentation like this for components will live in the storybook docs add-on

No

### DIFF
--- a/packages/discovery-react-components/src/components/SearchResults/__fixtures__/searchResults.ts
+++ b/packages/discovery-react-components/src/components/SearchResults/__fixtures__/searchResults.ts
@@ -12,7 +12,7 @@ const searchResults: DiscoveryV2.QueryResponse = {
       result_metadata: {
         collection_id: COLLECTION_ID_1
       },
-      title: 'alternate title',
+      title: 'Alternate title',
       document_passages: [
         {
           passage_text:
@@ -28,7 +28,7 @@ const searchResults: DiscoveryV2.QueryResponse = {
       result_metadata: {
         collection_id: COLLECTION_ID_1
       },
-      title: 'alternate title',
+      title: 'Alternate title',
       document_passages: [
         {
           passage_text:
@@ -47,7 +47,7 @@ const searchResults: DiscoveryV2.QueryResponse = {
       extracted_metadata: {
         title: 'IBM_Analytics_Machine_Learning.pdf'
       },
-      title: 'alternate title',
+      title: 'Alternate title',
       text:
         '<em>Machine learning</em> trains a model from patterns in your data, exploring a space of possible models defined by parameters. If your parameter space is too big, you’ll overfit to your training data and train a model that doesn’t generalize beyond it. A detailed explanation requires more math, but as a rule you should keep your models as simple as possible.'
     },
@@ -56,7 +56,7 @@ const searchResults: DiscoveryV2.QueryResponse = {
       result_metadata: {
         collection_id: COLLECTION_ID_1
       },
-      title: 'alternate title',
+      title: 'Alternate title',
       extracted_metadata: {
         title: 'IBM_Analytics_Machine_Learning.pdf'
       }
--- a/packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
+++ b/packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
@@ -113,6 +113,7 @@ export const Result: React.FunctionComponent<ResultProps> = ({
   if (result) {
     documentId = result.document_id;
   }
+
   const title = get(result, resultTitleField);
   const filename: string | undefined = get(result, 'extracted_metadata.filename');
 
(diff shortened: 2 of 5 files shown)


packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -113,7 +113,12 @@ export const Result: React.FunctionComponent<ResultProps> = ({
   if (result) {
     documentId = result.document_id;
   }
-  const title = get(result, resultTitleField);
+
+  let title = get(result, resultTitleField);
+  if (title && title.length > 100) {
+    title = title.substring(0, 100) + '...';
```
maniax89: i would use an ellipsis character here -> `…`

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -113,7 +113,12 @@ export const Result: React.FunctionComponent<ResultProps> = ({
   if (result) {
     documentId = result.document_id;
   }
-  const title = get(result, resultTitleField);
+
+  let title = get(result, resultTitleField);
+  if (title && title.length > 100) {
```
maniax89: is there a reason we decided on `100`? also would be good to enable a hover of the title to show the entire title if we are clipping it

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
maniax89: also is there any way to do this with CSS instead? `text-overflow: ellipsis` should accomplish what we need here

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
Steven-Gassert: 100 was an arbitrary choice that I felt would be able to contain most customers title choices. I'm open to feedback for a different value and agree that having a hover to display the full title would make for a better experience here

packages/discovery-styles/scss/components/search-results/_search-results.scss
```diff
@@ -171,15 +171,28 @@ $result-table-fade-offset: calc(#{$result-element-max-height} - #{$result-table-
   color: $gray-60;
   font-size: $spacing-sm;
   display: flex;
-  align-items: center;
```
Steven-Gassert: this is inside of `.#{$prefix}--search-result__footer` without aligning our collection and title name, it allows either one to grow to multiple lines in size, without the other one jumping down to stay aligned vertically

packages/discovery-styles/scss/components/search-results/_search-results.scss
```diff
@@ -171,15 +171,28 @@ $result-table-fade-offset: calc(#{$result-element-max-height} - #{$result-table-
...
+    white-space: normal;
+  }
+}
+
+.#{$prefix}--search-result__footer__collection-name {
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
 
-  svg {
```
Steven-Gassert: removing since we removed the icon at the beginning of https://github.ibm.com/Watson-Discovery/disco-widgets/issues/236

packages/discovery-react-components/src/components/SearchResults/__fixtures__/searchResults.ts
```diff
@@ -12,7 +12,7 @@ const searchResults: DiscoveryV2.QueryResponse = {
       result_metadata: {
         collection_id: COLLECTION_ID_1
       },
-      title: 'alternate title',
+      title: 'Alternate title',
```
Steven-Gassert: This just looks nicer in storybook

packages/discovery-styles/scss/components/search-results/_search-results.scss
```diff
@@ -171,15 +171,28 @@ $result-table-fade-offset: calc(#{$result-element-max-height} - #{$result-table-
...
 
 .#{$prefix}--search-result__footer__title {
-  align-items: center;
-  display: flex;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  margin-right: $spacing-05;
+
+  &:hover {
```
maniax89: might want to check with design on this one, the search results moving around on hover might be a little jarring

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -113,6 +113,7 @@ export const Result: React.FunctionComponent<ResultProps> = ({
   if (result) {
     documentId = result.document_id;
   }
+
```
maniax89: another option for hover is to just pass the `title=` attribute to the various divs below and the tooltip on hover will show the entire thing (based on your OS/browser)

PR #10, opened 2019-12-03
chore: Update dependencies

ChrisGillin:

Update package dependencies

Run the example app and make sure existing functionality is unaffected.

No

### DIFF
--- a/packages/discovery-react-components/package.json
+++ b/packages/discovery-react-components/package.json
@@ -26,13 +26,13 @@
     "dist"
   ],
   "peerDependencies": {
-    "@carbon/icons": "^10.5.0",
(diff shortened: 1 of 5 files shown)


packages/discovery-react-components/package.json
```diff
@@ -26,13 +26,13 @@
     "dist"
   ],
   "peerDependencies": {
-    "@carbon/icons": "^10.5.0",
+    "@carbon/icons": "^10.6.1",
```
maniax89: for peer dependencies in general, we should probably specify minimum versions for our components to work https://nodejs.org/es/blog/npm/peer-dependencies/

things like `10.x.x` or `>= 10.5.0 < 11`

PR #11, opened 2019-12-03
bug: Clear all button working with collection facet

kabeaty:
Previously, the root `Clear all` button for the SearchFacets component was not appearing when collectionFacets were selected, and it was also not clearing collectionFacets when it was clicked. These changes make it so that this `Clear all` button interacts with the collectionFacets component as it does the other facets components.

related to issues: 
https://github.ibm.com/watson-discovery/disco-widgets/issues/427
https://github.ibm.com/watson-discovery/disco-widgets/issues/428

I added some tests for this interaction.
You can also test in the example app with a project that has collections. Select a collection and see that the `Clear all` button appears. Click the `Clear all` button and see that it deselects the collections that you selected.

I added comments to props that were added, and I also left a comment in the SearchFacets component regarding what we should change when Carbon MultiSelect offers more control over selection. I also opened a Carbon enhancement request here: https://github.com/carbon-design-system/carbon/issues/4812.

I don't believe so since new props are only being added to the CollectionFacets component, but not to the top-level SearchFacets component.


packages/discovery-react-components/src/components/SearchFacets/cssClasses.ts
```diff
@@ -8,3 +8,5 @@ export const optionClass = `${baseClass}__facet__option`;
 export const optionLabelClass = `${baseClass}__facet__option-label`;
 export const singleSelectGroupClass = `${baseClass}__facet__single__select__group`;
 export const labelAndSelectionContainerClass = `${baseClass}__facet__label-and-selection-container`;
+
+export const collectionFacetIdPrefix = 'collection-facet-';
```
maniax89: not sure this is a css class, but no big deal

PR #13, opened 2019-12-04
Clone document input JSON into new object to maintain the immutabilit…

ColbyJohnIBM:
…y of the input

No longer add any data to the document input JSON object.

https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/195
https://github.ibm.com/Watson-Discovery/wd-issues/issues/1365

Linked this repo up to `discovery-tooling` and saw that these changes fixed the issue noted above (and in the linked issues).

N/A

No


anibapat:
It actually removes `metadata` from the metadata panel. Also When switching to Invoices/POs, it goes into infinite loop.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -99,7 +100,8 @@ const INITIAL_STATE: State = {
   isError: false,
   styles: [],
   sections: [],
-  itemMap: { byItem: {}, bySection: {} }
+  itemMap: { byItem: {}, bySection: {} },
+  metadata: []
```
anibapat: I think we need to do similar for `attributes` and `relations` since we add that as part of the response json. Those two properties are required in the Invoices/POs

JuanNicolasGarcia:
Screen Shot 2019-12-09 at 3 27 15 PM

Verified that the metadata field is not present in enriched_html field.

anibapat:
@maniax89 Can we get a review?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -34,7 +34,16 @@ import {
...
+import {
+  Metadata,
+  MetadataData,
+  EnrichedHtml,
+  Contract,
+  Item,
+  Field,
+  Attributes,
+  Relations
+} from '../../types';
```
maniax89: do we need to do `../../` here? i thought that `components/CIDocument/types` was more readable

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -67,7 +71,7 @@ export interface Table {
  * Convert document data into structure that is more palatable for use by
  * CIDocument
  *
- * @param {Object} data Discovery document data
+ * @param {Object} query Discovery document data
```
jhpedemonte: I don't like this change. `query` implies something else. If `data` didn't work, then perhaps `queryResult` or `queryData`.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -342,7 +364,7 @@ const CIDocument: FC<CIDocumentProps> = ({
             <Tab tabIndex={0} label={messages.metadataTabLabel}>
               {!hasError && (
                 <MetadataPane
-                  metadata={metadata}
+                  metadata={state.metadata}
```
jhpedemonte: Issue mentions that we need to fix this for `metadata` (doing that here) and `attributes` and `relations` for invoices/POs -- where are we doing the latter?

packages/discovery-react-components/src/utils/document/processDoc.ts
anibapat: queryData it is 👍

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
anibapat: https://github.com/watson-developer-cloud/discovery-components/pull/13/files#diff-158d0c756b5100722b593c523edcbd52R182 This change takes care or getting attribute / relation from the `state`

packages/discovery-react-components/src/utils/document/processDoc.ts
anibapat: Fixed https://github.com/watson-developer-cloud/discovery-components/pull/13/commits/f7415ec528350185a2ea54d0dfe82c468f2a796e

packages/discovery-react-components/src/utils/document/__tests__/processDoc.spec.tsx
```diff
@@ -31,6 +33,19 @@ describe('processDoc', () => {
     doc = await processDoc(contractData.results[0], { sections: true });
   });
 
+  it('does not mutate the passed document data and metadata property are not be present in it', () => {
+    expect(doc.metadata).toHaveLength(4);
+    expect(get(contractData.results[0], 'enriched_html[0].contract.metadata')).toBeUndefined();
```
jhpedemonte: I would prefer setting up a test that did a deep clone of the original data, then doing deep compare of original data against clone. That way we can ensure that _nothing_ is changed, not just `metadata`.

packages/discovery-react-components/src/utils/document/__tests__/processDoc.spec.tsx
```diff
@@ -31,6 +33,19 @@ describe('processDoc', () => {
...
+  });
+
+  it('does not mutate the passed document data and attributes and relations properties are not present in it', async () => {
+    const invoiceDoc = await processDoc(invoiceData, { sections: true });
+    expect(invoiceDoc.attributes).toHaveLength(5);
+    expect(invoiceDoc.relations).toHaveLength(6);
+    expect(get(invoiceData, 'enriched_html[0].invoice.attributes')).toBeUndefined();
+    expect(get(invoiceData, 'enriched_html[0].invoice.relations')).toBeUndefined();
+  });
```
jhpedemonte: Same with this test, just passing in invoice data.

packages/discovery-react-components/src/utils/document/__tests__/processDoc.spec.tsx
anibapat: That will just check if `lodash clonedeep` did  the job or not,  `processDoc` function already has a clonedeep which clones  the original data, make some transformations to it and return  `data` which has only 4 props.  
```
{styles: "/**/↵ .Arial_BoldMT_Black_ }↵/**/"
metadata: (2) [{…}, {…}]
sections: (237) [{…}, {…}, …]
itemMap: {byItem: {…}, bySection: {…}}

So comparing original data and returned data would always have a difference so I think it is necessary to check explicitly if original data does not have `metadata` / `attributes`, `relations` after calling `processDoc` function (which already has a deepcloned original data). Thoughts?

jhpedemonte: Right, but wasn't the original issue that we **weren't** doing a deep clone in `processDoc` and instead updating the incoming document data object?

anibapat: Right! We weren't doing a cloneDeep earlier, So now that `processDoc` has it, input data can be checked after running a `processDoc` function, and then check explicitly if `metadata` field is not present in the original input data.

jhpedemonte: What I don't like is that this test is too specific -- it only checks if we are adding `metadata`. But if future changes pollute the incoming document data with some other property? I want a general test that would catch such a bug.

anibapat: Done! Added another check for entire data. Also kept metadata check since this specific issue is fixed in this PR.

PR #15, opened 2019-12-05
chore: Add module path aliases

kabeaty:
This adds in module path aliases to replace the complex relative `../../` imports throughout `discovery-react-components`.

To get the aliases to work with Storybook, I had to include a different `baseUrl` in its `tsconfig.json` and use a `tsconfig-paths-webpack-plugin` in its Webpack config.

I also added in [`madge`](https://github.com/pahen/madge) to help us identify any circular dependencies. I added just the circular-dependencies check for now, but we could set up `madge` to do more in the future too.  I made this a separate command so we can run it ad hoc, and then it's also included in our lint.

I did identify 3 circular dependencies with `madge` that I also fixed in this PR.

related issue: https://github.ibm.com/Watson-Discovery/disco-widgets/issues/278

You can verify these changes by checking that builds, tests, Storybook, and the example app all still work as expected with no errors.

Note that you may need to restart VSCode to get it to recognize the module path aliases.

Yes, I've updated both the `README` and `CONTRIBUTING` with information about the module path aliases for imports and the new circular-dependencies command.

These shouldn't be breaking changes.


packages/discovery-react-components/.storybook/tsconfig.json
```diff
@@ -1,6 +1,17 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "isolatedModules": false
+    "isolatedModules": false,
+    "baseUrl": "../src",
+    "paths": {
+      "@rootUtils/*": ["utils/*"],
```
maniax89: can we just do `utils`?

packages/discovery-react-components/src/components/CIDocument/CIDocument.tsx
```diff
@@ -2,13 +2,13 @@ import CIDocument, {
   CIDocumentProps,
   canRenderCIDocument,
   Messages
-} from './components/CIDocument/CIDocument';
+} from '@CIDocument/components/CIDocument/CIDocument';
```
maniax89: i was hoping to just address `../../` situations. this seems more verbose than before

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -16,49 +16,48 @@ import get from 'lodash/get';
...
-import MetadataPane from '../MetadataPane/MetadataPane';
-import NavigationToolbar from '../NavigationToolbar/NavigationToolbar';
-import CIDocumentContent from '../CIDocumentContent/CIDocumentContent';
-import processDoc from '../../../../utils/document/processDoc';
-import { getEnrichmentName } from '../../utils/enrichmentUtils';
-import { getDetailsFromItem, getDetailsFromMetadata } from '../../utils/details';
-import { getFilterHelper, ProcessFilter } from '../../utils/filterHelper';
-import { isFilterEmpty, updateFilter } from '../../utils/filterUtils';
-import { getId, findElement, findElementIndex } from '../../../../utils/document/idUtils';
+import DetailsPane from '@CIDocument/components/DetailsPane/DetailsPane';
```
maniax89: how do we feel about `components/CIDocument` instead of `@CIDocument`?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -16,49 +16,48 @@ import get from 'lodash/get';
...
+import FilterPanel from '@CIDocument/components/FilterPanel/FilterPanel';
+import MetadataPane from '@CIDocument/components/MetadataPane/MetadataPane';
+import NavigationToolbar from '@CIDocument/components/NavigationToolbar/NavigationToolbar';
+import CIDocumentContent from '@CIDocument/components/CIDocumentContent/CIDocumentContent';
+import processDoc from '@rootUtils/document/processDoc';
```
maniax89: could we just do `utils`? also not sure if its for this PR but i wonder if doing something like `import { processDoc } from 'utils/document';` reads easier

packages/discovery-react-components/tsconfig.prod.json
```diff
@@ -2,6 +2,17 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "declaration": true,
-    "isolatedModules": false
+    "isolatedModules": false,
+    "baseUrl": "src",
+    "paths": {
```
maniax89: can we put this in the `tsconfig.json` instead? or does it only work here?

packages/discovery-react-components/package.json
```diff
@@ -11,6 +11,7 @@
     "test": "cross-env CI=1 react-scripts test",
     "test:watch": "react-scripts test --watch",
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
+    "circular-dependencies": "madge --circular src/*",
```
maniax89: can we add this as part of our CI build? maybe just add it as part of the `lint` task

packages/discovery-react-components/.storybook/tsconfig.json
kabeaty: I'm open to it. My thought here was that this clearly distinguished it from the `utils` we have inside of some of our components folders.

packages/discovery-react-components/src/components/CIDocument/CIDocument.tsx
kabeaty: Fair. I also thought about adding aliases for nested components, not just top-level ones. I think that would keep us consistently using module path aliases but reduce the verbosity. What are your thoughts on that change?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
kabeaty: I'm open to using that convention instead if we don't feel we need a prefix.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
kabeaty: See my note above on why I went with `rootUtils` (to distinguish from the utils nested in component folders). I agree that updates like that would make imports read easier, and I think it could make sense to do in this PR, so I'll go through and do that :)

packages/discovery-react-components/tsconfig.prod.json
kabeaty: Yeah, I mentioned this in the PR description too, but if the `paths` are in the `tsconfig.json`, they get removed by `react-scripts` when running our tests. Hence placing them here and then also having to define them in our Storybook's tsconfig.

packages/discovery-react-components/package.json
kabeaty: Yeah, I mentioned in the PR description that I was considering putting in our build, although wasn't sure if we'd want it every time and increase the time the build takes. I think putting it in the `lint` task could be a good way to make sure it's run and checked with doing it every build though, I like that idea 👍

package.json
```diff
@@ -76,11 +77,12 @@
     "rollup-plugin-url": "^2.2.2",
     "source-map-explorer": "2.0.1",
     "start-server-and-test": "^1.10.0",
+    "tsconfig-paths-webpack-plugin": "^3.2.0",
     "typescript": "^3.5.3",
     "webpack": "4.39.1"
   },
   "scripts": {
-    "lint": "eslint '{packages,examples}/**/*.{js,jsx,ts,tsx}' --ignore-pattern node_modules --quiet",
+    "lint": "eslint '{packages,examples}/**/*.{js,jsx,ts,tsx}' --ignore-pattern node_modules --quiet && madge --circular packages/discovery-react-components/src/*",
```
maniax89: I think it would be nice to have this as separate methods like 
```
"lint": "yarn run eslint && yarn run circular"

Comment (maniax89) [reply package.json]: can we define a `lint` command in our `discovery-react-components` package? i think running `yarn lint` will automatically run that as well

File: package.json [hunk 8]
```diff
@@ -76,11 +77,12 @@
...
     "lint:fix": "yarn run lint --fix"
```
maniax89: might need to remove this because of how `lint` is defined

PR #16, opened 2019-12-06
chore: Update carbon-components-react to at least 7.7.0

kabeaty:
Upgrade `carbon-components-react` to at least 7.7.0 so that we no longer see `console.error ../../node_modules/prop-types/factoryWithTypeCheckers.js:23
      Warning: Invalid arguments supplied to oneOf, expected an array, got 4 arguments. A common mistake is to write oneOf(x, y, z) instead of oneOf([x, y, z]).` when running our tests.

Run `yarn test` and see that this error is no longer there. Also check that everything else still works as expected (build, storybook, etc.)

n/a

No


packages/discovery-react-components/package.json
```diff
@@ -29,7 +29,7 @@
     "@carbon/icons": ">= 10.5.0 < 11",
     "@ibm-watson/discovery-styles": "1.x.x",
     "carbon-components": ">= 10.6.0 < 11",
-    "carbon-components-react": ">= 7.6.0 < 8",
+    "carbon-components-react": ">= 7.7.0 < 8",
```
ChrisGillin: If this is just to avoid some warnings when running tests do we really want to update the peer dependency too? Can't we just update the lock file?

maniax89: I agree, we don't necessarily need 7.7 to run, but locally to get rid of errors we can bump.

Peer dependencies in general should be lenient https://nodejs.org/es/blog/npm/peer-dependencies/

PR #17, opened 2019-12-06
chore(eslint): disable formatOnSave and add jsx-a11y

Steven-Gassert:
This disables the formatOnSave hook as well as splits up our lint rules for js,jsx,ts,tsx and other files in our repo. This also enables the `jsx-a11y` plugin for issue https://github.ibm.com/Watson-Discovery/disco-widgets/issues/166 and https://github.ibm.com/Watson-Discovery/disco-widgets/issues/168 . I'm opening this as a draft PR to try to get some feedback from the doc-viz team. When enabling the `jsx-a11y` plugin, the `Section.tsx` component contains some violations. Feedback from the owners of this component would help me more confidently tackle these violations as the fix requires adding event listeners ect (more detail contained in the diff comments)

- Make sure that the formatOnSave hook does not run
- Make sure that prettier is run on the pre-commit hook
- Make sure that the a11y plugin is added and that there are no violations

N/A

no


.eslintrc
```diff
@@ -3,12 +3,13 @@
   "extends": [
     "plugin:react/recommended",
     "plugin:@typescript-eslint/recommended",
-    "plugin:prettier/recommended"
+    "plugin:prettier/recommended",
+    "plugin:jsx-a11y/recommended"
```
Steven-Gassert: @jhpedemonte seemed to have utilized the recommended settings before. When checking to see which settings CRA uses by default (the root of why this issue was opened) I'm unsure whether these would conflict, see what I presume to be the CRA defaults [here](https://github.com/facebook/create-react-app/blob/master/packages/eslint-config-react-app/index.js#L285)

This made me question whether we should just attempt to use this [package](https://github.com/facebook/create-react-app/tree/master/packages/eslint-config-react-app) at the root of our repository to make sure that we are using CRA defaults across the board and don't see similar CI/CD failures

package.json
```diff
@@ -82,9 +83,13 @@
     "lint:fix": "yarn run lint --fix"
   },
   "lint-staged": {
-    "{packages,examples}/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
+    "{packages,examples}/**/*.{json,css,scss,md}": [
       "prettier --write",
       "git add"
+    ],
+    "{packages,examples}/**/*.js,jsx,ts,tsx": [
```
Steven-Gassert: I split these two file types [per Javiers suggestion](https://github.ibm.com/Watson-Discovery/disco-widgets/issues/166#issuecomment-15076024)

packages/discovery-react-components/src/components/CIDocument/components/Section/Section.tsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
Steven-Gassert: I have some guesses as to what this component does, but currently our cluster is down and it's preventing me from being confident in the exact functionality. This rule, and the one below it, are the two `a11y` violations in this repo

`click-events-have-key-events` states that anything you can click, you should also be able to navigate with the keyboard and select. the only problem is that the onClick handler `mouseClickListener` takes a `mouseEvent` and not a `keyboardEvent`. I'm not able to say that this function can take a `mouseEvent` _or_ `keyboardEvent` because this handler makes use or mouse positioning `client.X` and `client.Y` and I'm unsure of just using the current cursor position is ideal in this scenario and would love some feedback as to what might be the best approach here (if not disabling the rule)

`no-static-element-interactions` states that you need to provide a `role` to this element. the list of options can be found [here](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md#case-this-element-acts-like-a-button-link-menuitem-etc). I originally set the `role` to `button`, but in that case you need to make choices about what the tab-ability should be, in this case some feedback would also be helpful (if not to disable this rule too)

package.json
```diff
@@ -84,9 +85,13 @@
...
+      "eslint --fix",
```
maniax89: What happens if eslint cannot fix these? Usually prettier only includes stylistic changes whereas eslint can make other changes. Also we miss out on getting the prettier changes.

Do we want to look into https://github.com/prettier/prettier-eslint instead?

.vscode/settings.json
```diff
@@ -12,7 +12,6 @@
       "autoFix": true
     }
   ],
-  "editor.formatOnSave": true,
```
maniax89: If you are going to remove this, you might as we remove the whole file.

Some good reads before doing so:
https://glebbahmutov.com/blog/configure-prettier-in-vscode/
https://levelup.gitconnected.com/setting-up-eslint-with-prettier-typescript-and-visual-studio-code-d113bbec9857

package.json
Steven-Gassert: I'm going to combine these two file types again, although `eslint` will have no effect on css and scss, this will still invoke prettier which should have an effect

packages/discovery-react-components/src/components/CIDocument/components/Section/Section.tsx
Steven-Gassert: after adding in `eslint-config-react-app`, these accessibility rules no longer apply

.vscode/settings.json
Steven-Gassert: After chatting, we've decided that since our linting rules no longer conflict, we will keep `formatOnSave` functionality

Steven-Gassert:
closing in favor of https://github.com/watson-developer-cloud/discovery-components/pull/19

PR #19, opened 2019-12-10
chore(lint): use eslint-config-react-app

Steven-Gassert:
This updates our linter to use `eslint-config-react-app`. This also explicitly runs `eslint` on all `.ts`,`.tsx`, `.js`, and `.jsx` files and `prettier` on all other file types. This is related to https://github.ibm.com/Watson-Discovery/disco-widgets/issues/166 and https://github.ibm.com/Watson-Discovery/disco-widgets/issues/168

for issue https://github.ibm.com/Watson-Discovery/disco-widgets/issues/168 , note [here](https://www.npmjs.com/package/eslint-config-react-app#accessibility-checks) that `eslint-config-react-app` contains the same accessibility checks that CRA does.

for issue https://github.ibm.com/Watson-Discovery/disco-widgets/issues/168 note that our eslint and prettier plugins are configured correctly and that you do not see alternating eslint errors when saving code in VSCode

N/A
no


package.json
```diff
@@ -29,12 +29,12 @@
     "@types/react-dom": "^16.9.0",
     "@types/react-virtualized": "^9.21.2",
     "@types/seedrandom": "^2.4.28",
-    "@types/storybook__addon-knobs": "^5.0.3",
```
Steven-Gassert: note that the latest version of storybook provides it's own types https://www.npmjs.com/package/@types/storybook__addon-knobs

package.json
```diff
@@ -84,9 +87,13 @@
     "lint:fix": "yarn run lint --fix"
   },
   "lint-staged": {
-    "{packages,examples}/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
+    "{packages,examples}/**/*.{json,css,scss,md}": [
```
Steven-Gassert: I ended up keeping these rules more explicitly split. It turns out that eslint would have no effect on `json,css,scss,md` files anyway. Splitting the definitions I felt made it more clear as to what packages are being run on which files

package.json
```diff
@@ -43,7 +43,10 @@
     "cypress": "3.2.0",
     "eslint": "^6.3.0",
     "eslint-config-prettier": "^6.2.0",
+    "eslint-config-react-app": "^5.1.0",
+    "eslint-plugin-flowtype": "^4.5.2",
     "eslint-plugin-import": "^2.18.2",
+    "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-prettier": "^3.1.0",
```
maniax89: can we remove any of these explicit dependencies now? let's make sure anything removed from eslintrc are also removed from our dev dependencies

Steven-Gassert: I double checked over our `plugins`, and it looks like everything we explicitly removed from `.eslintrc` was also required as a peer dependency to get `eslint-config-react-app` to start

PR #22, opened 2019-12-13
feat: structured query basic

kabeaty:
These changes add in the basic `StructuredQuery` component, placing everything on the page and setting up tests, Storybook, and basic styles, too.

You can check out the tests that I added for the new components. You can also take a look at the default story that I added for the `StructuredQuery` component in Storybook.

I have added Storybook to begin the documentation for the `StructuredQuery` component.

No, since this is a new component addition, it doesn't include breaking changes.


packages/discovery-react-components/src/components/StructuredQuery/__stories__/default.md
```diff
@@ -0,0 +1,3 @@
+#### Overview
+
+The `StructuredQuery` component will provide the ability to visually build structured queries against your project. Full functionality will be added in the future.
```
maniax89: probably don't need to say anything about the future

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export interface RuleRowProps {
+  /**
+   * override default messages for the component by specifying custom and/or internationalized text strings
+   */
+  messages: Messages;
+}
+
+export const RuleRow: React.FunctionComponent<RuleRowProps> = ({ messages }) => {
+  const handleOnFieldChange = () => {
```
maniax89: probably can remove this stuff for this PR

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div className={structuredQueryRulesClass}>
+      <ComboBox
+        id="structured-query-rules-field-0"
+        // TODO: Items is empty for now as it's a required field and retrieving fields for the dropdown
+        // and adding them as items will be addressed in a future issue
+        items={[]}
+        placeholder={messages.fieldDropdownPlaceholderText}
+        titleText={messages.fieldDropdownTitleText}
+        onChange={handleOnFieldChange}
```
maniax89: don't need to include these until we implement them

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  const handleOnOperatorChange = () => {
+    // TODO: Fully implement handling operator dropdown selections in a future issue
+  };
+
+  return (
+    <div className={structuredQueryRulesClass}>
+      <ComboBox
+        id="structured-query-rules-field-0"
```
maniax89: for this ID should we be passing it down or randomizing it? what happens if it isn't provided? this would potentially create issues if more than one `RuleRow` is on the page (which im imagining there will be)

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+        // TODO: Items is empty for now as it's a required field and retrieving fields for the dropdown
+        // and adding them as items will be addressed in a future issue
+        items={[]}
+        placeholder={messages.fieldDropdownPlaceholderText}
+        titleText={messages.fieldDropdownTitleText}
+        onChange={handleOnFieldChange}
+      />
+      <ComboBox
+        id="structured-query-rules-operator-0"
```
maniax89: same here - should we be passing this ID in? let it default to something? or add some randomized string to it so it doesn't conflict with other `<RuleRow />` components on the page

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+        items={[
+          messages.operatorDropdownIsOptionText,
```
maniax89: I think these will need to have values in addition to their text so that regardless of what is set as the `message` we receive the same programatic value for each

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+          messages.operatorDropdownIsNotOptionText,
+          messages.operatorDropdownContainsOptionText,
+          messages.operatorDropdownDoesNotContainOptionText
+        ]}
+        placeholder={messages.operatorDropdownPlaceholderText}
+        titleText={messages.operatorDropdownTitleText}
+        onChange={handleOnOperatorChange}
```
maniax89: probably can remove this

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+      />
+      <TextInput
+        id="structured-query-rules-value-0"
```
maniax89: same comment about IDs here

packages/discovery-react-components/src/components/StructuredQuery/components/SatisfyRules/SatisfyRules.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const SatisfyRules: React.FunctionComponent<SatisfyRulesProps> = ({ messages }) => {
+  const handleOnChange = () => {
+    // TODO: Fully implement handling satisfy rules dropdown selections in a future issue
+  };
+
+  const satisfyRulesDropdownElement = (
+    <Dropdown
+      id="structured-query-dropdown"
```
maniax89: same comment about ids here

packages/discovery-react-components/src/components/StructuredQuery/components/SatisfyRules/SatisfyRules.tsx
```diff
@@ -0,0 +1,42 @@
...
+      items={[
+        messages.satisfyRulesDropdownAllOptionText,
```
maniax89: same comment about messages/values here

packages/discovery-react-components/src/components/StructuredQuery/components/SatisfyRules/SatisfyRules.tsx
```diff
@@ -0,0 +1,42 @@
...
+        messages.satisfyRulesDropdownAnyOptionText
+      ]}
+      type="inline"
+      initialSelectedItem={messages.satisfyRulesDropdownAllOptionText}
+      label={messages.satisfyRulesDropdownLabelText}
+      onChange={handleOnChange}
+      key="structured-query-dropdown"
```
maniax89: i don't think this `key` is needed

packages/discovery-react-components/src/components/StructuredQuery/cssClasses.ts
```diff
@@ -0,0 +1,6 @@
+import { settings } from 'carbon-components';
+
+export const structuredQueryClass = `${settings.prefix}--structured-query`;
+export const structuredQuerySatisfyRulesDropdownClass = `${settings.prefix}--structured-query__satisfy-rules-dropdown`;
```
maniax89: you could build this off of `structuredQueryClass` -> `${structuredQueryClass}__satisfy-rules-dropdown`

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
```diff
@@ -0,0 +1,86 @@
...
+   */
+  addRuleText: string;
+  /**
+   * override the default text for the button to add a group of rules
+   */
+  addGroupRulesText: string;
+}
+
+export const defaultMessages: Messages = {
+  satisfyRulesDropdownText: 'Satisfy %dropdown% of the following rules',
```
maniax89: i think the syntax that other libraries use for variable interpolation is `{dropdown}` -> 
1. i18n next -> https://www.i18next.com/translation-function/interpolation
1. icu message syntax -> https://formatjs.io/guides/message-syntax/

packages/discovery-react-components/src/components/StructuredQuery/utils/formatSatisfyRulesDropdownMessage.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+
+export const formatSatisfyRulesDropdownMessage = (message: string, dropdown: JSX.Element) => {
```
maniax89: i think this will be a useful utility across our components that we can make a bit more generic

something like

```suggestion
export const formatSatisfyRulesDropdownMessage = (message: string, dropdown: JSX.Element) => {

interface SubstitutionValues {
  [name: string]: React.Node // this includes strings as well as JSX elements i think
}

const SPLIT_VARIABLES_REGEX = /({[^}]+})/;
formatMessage(message: string, values: SubstitutionValues) => {
  const variables = Object.keys(values);
  return message.split(SPLIT_VARIABLES_REGEX).map((part, i) => {
   const variableWithoutBraces = part.replace(/{}/g, '');
   return part.contains('{') ? <div key={i}>{values[variableWithoutBraces]}</div> : part;
  });
}
```

packages/discovery-react-components/src/components/StructuredQuery/__stories__/default.md
kabeaty: Okay, either way makes sense to me. So, just to confirm on the same page, you're thinking no overview at all for now and we add it in later?

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
kabeaty: I added these in now since `onChange` is a required prop for `ComboBox` and I didn't want to see a warning in Storybook.

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
kabeaty: Yeah, I'm handling this more in the next issue where we actually have multiple `RuleRow`s on the page. If it's not provided, we get a warning since it's required. Basically, my implementation in the next issue so far is to use the `id` of the rule row, which will be part of what's passed down. I figured it was okay to leave this way at this point since there was no way to add another rule row so far.

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
kabeaty: See my note above on the same :)

packages/discovery-react-components/src/components/StructuredQuery/components/SatisfyRules/SatisfyRules.tsx
kabeaty: I added this since we get a warning if we don't have the key here that we need it.

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
kabeaty: All right, will update to use the brackets 👍

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
kabeaty: Yeah, makes sense, I can add these values in now.

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
maniax89: we can call this out in our contributing guidelines as well

packages/discovery-react-components/src/utils/__tests__/formatMessage.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+describe('formatMessage util', () => {
+  test('correctly formats a message when provided with a message and elements to interpolate', () => {
+    const messageToFormat = 'Format me with a {button} and a {link}';
+    const button = <button>I am button</button>;
+    const link = <a>I am link</a>;
+    const formattedMessage = render(
+      <div>{formatMessage(messageToFormat, { button: button, link: link })}</div>
+    );
+    const buttonMatch = formattedMessage.getByText('I am button');
```
maniax89: does it work when you do `formattedMessage.getByText('Format me with a I am button and a I am link')`?

packages/discovery-react-components/src/utils/formatMessage.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
```
maniax89: i know we had been recommending doing `import * from React` before for storybook, but I believe we don't need to do that here (or anywhere) as long as we don't mix `React.Thing` and just `Thing`

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
```diff
@@ -0,0 +1,86 @@
...
+   */
+  addRuleText: string;
+  /**
+   * override the default text for the button to add a group of rules
+   */
+  addGroupRulesText: string;
+}
+
+export const defaultMessages: Messages = {
+  satisfyRulesDropdownText: 'Satisfy {dropdown} of the following rules',
```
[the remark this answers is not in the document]
kabeaty: Yeah, I think that makes sense since it's really kind of the header for a rule group. Will update, thanks!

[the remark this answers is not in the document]
kabeaty: Might also make sense to rename the `SatisfyRules` component to something like `RuleGroupDropdown` in this case to avoid any confusion of referring to them with two different names?

packages/discovery-react-components/src/utils/__tests__/formatMessage.test.tsx
kabeaty: No, since the text is broken up across elements, this doesn't work...

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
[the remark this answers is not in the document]
maniax89: sure!

packages/discovery-react-components/src/utils/formatMessage.tsx
kabeaty: Ah yes, this snuck back in from before. I can also add something to CONTRIBUTING and confirm we don't have any lingering elsewhere.

PR #23, opened 2019-12-17
chore: add error boundary to all components

Steven-Gassert:
Adds `react-error-boundary` to all components except for `DocumentPreview` as well as `CIDocument` since they were already using the custom utility `withErrorBoundary`. If we are interested in moving these components onto `react-error-boundary` for the sake of consistency I can migrate these components as well

to verify these changes I added `throw new Error('This is a new error');` inside of our components to test the fallback component they rendered. additionally I checked that the error and component stack were logged in the console. See example below
Screen Shot 2019-12-16 at 9 12 22 PM

N/A
no


.vscode/settings.json
```diff
@@ -1,17 +1,4 @@
 {
-  "eslint.autoFixOnSave": true,
```
Steven-Gassert: This was automatically converted by the eslint plugin in my code editor. I checked to make sure that autosave still functioned properly after these changes were made

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -246,7 +249,7 @@ export const DiscoverySearch: FC<DiscoverySearchProps> = ({
         }
       });
     },
-    [performSearch, setSearchParameters]
+    [performSearch, searchClient, setSearchParameters]
```
Steven-Gassert: I've removed this previously but I believe it keeps getting added in when we autofix with eslint before commit

packages/discovery-react-components/src/utils/onErrorCallback.ts
```diff
@@ -0,0 +1,5 @@
+const onErrorCallback = (error: Error, componentStack: string): void => {
+  console.error('Error caught', error, componentStack);
```
maniax89: im surprised you dont need an eslint rule. also what exactly does this look like?

.vscode/settings.json
Steven-Gassert: Yup, looks like this comes with a vscode upgrade which deprecates the `eslint.autoFixOnSave` option https://code.visualstudio.com/updates/v1_41#_eslint

One change I did push up here is to move `editor.codeActionsOnSave` so that the linter rules come before the formatting rules

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -440,3 +443,9 @@ export const DiscoverySearch: FC<DiscoverySearchProps> = ({
     </SearchApi.Provider>
   );
 };
+
+export default withErrorBoundary(
+  DiscoverySearch,
+  FallbackComponent('root discovery search provider '),
```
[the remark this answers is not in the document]
Steven-Gassert: 👍 I've changed each one of these messages to be the component name
(ie `There was an error rendering DiscoverySearch`)

Steven-Gassert:
@maniax89 I considered writing a test for this but felt like in this instance the tests inside of react-error-boundary covered the use cases I could think of, ex. see here https://github.com/bvaughn/react-error-boundary/blob/master/src/__tests__/ErrorBoundary.test.js if it makes sense to also test this within our repo, maybe by rendering a component which always throws an error than I don't mind including this

As far as the logging the componentStack and Error to the console, see this screenshot
Screen Shot 2019-12-17 at 3 18 26 PM

packages/discovery-react-components/src/components/SearchInput/SearchInput.tsx
```diff
@@ -285,4 +288,4 @@ export const SearchInput: FC<SearchInputProps> = props => {
   );
 };
 
-export default SearchInput;
+export default withErrorBoundary(SearchInput, FallbackComponent('SearchInpur'), onErrorCallback);
```
kabeaty: `SearchInpur` --> `SearchInput` :)

packages/discovery-react-components/src/index.tsx
```diff
@@ -1,12 +1,12 @@
...
 } from './components/DiscoverySearch/DiscoverySearch';
 export { SearchInput } from './components/SearchInput/SearchInput';
-export { SearchFacets } from './components/SearchFacets/SearchFacets';
+export { default as SearchFacets } from './components/SearchFacets/SearchFacets';
 export { ResultsPagination } from './components/ResultsPagination/ResultsPagination';
 export { DocumentPreview } from './components/DocumentPreview/DocumentPreview';
-export { SearchResults } from './components/SearchResults/SearchResults';
+export { default as SearchResults } from './components/SearchResults/SearchResults';
 export { CIDocument, canRenderCIDocument } from './components/CIDocument/CIDocument';
 export { StructuredQuery } from './components/StructuredQuery/StructuredQuery';
```
maniax89: can you add to `StructuredQuery` as well?

PR #27, opened 2019-12-19
feat: structured query add and remove rules

kabeaty:
This adds in the ability to add and remove rule rows from rule rows in the `StructuredQuery` component. When clicking the `Add rule` button, you should see a new row of rules added. When there is more than one rule row, a `Remove row` icon should appear on the right side of the rule row. When that is clicked, the row should be removed. I set up a max number of allowed sibling rule rows, which is set to 3 for now, pending user feedback.

related to issue: https://github.ibm.com/watson-discovery/disco-widgets/issues/182

I added some tests for these new additions, so you can look over those and confirm they are fully and correctly testing these new features. You can also run Storybook and add/remove rules in the `StructuredQuery` default story there.

`StructuredQuery` is in Storybook, but nothing more was added there specifically for these features.

no


packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -1,44 +1,81 @@
...
+        rows: ruleRows.rows.filter(ruleRow => ruleRow.id !== rowId)
+      })
+    );
+  };
 
   return (
     <div className={structuredQueryRulesClass}>
       <ComboBox
-        id="structured-query-rules-field-0"
+        id={`structured-query-rules-field-` + rowId}
```
maniax89:
```suggestion
        id={`structured-query-rules-field-${rowId}`}
```

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -1,44 +1,81 @@
...
         // TODO: Items is empty for now as it's a required field and retrieving fields for the dropdown
         // and adding them as items will be addressed in a future issue
         items={[]}
         placeholder={messages.fieldDropdownPlaceholderText}
         titleText={messages.fieldDropdownTitleText}
       />
       <ComboBox
-        id="structured-query-rules-operator-0"
+        id={`structured-query-rules-operator-` + rowId}
```
maniax89:
```suggestion
        id={`structured-query-rules-operator-${rowId}`}
```

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -1,44 +1,81 @@
...
         items={operatorDropdownItems}
         placeholder={messages.operatorDropdownPlaceholderText}
         titleText={messages.operatorDropdownTitleText}
       />
       <TextInput
-        id="structured-query-rules-value-0"
+        id={`structured-query-rules-value-` + rowId}
```
maniax89:
```suggestion
        id={`structured-query-rules-value-${rowId}`}
```

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -1,44 +1,81 @@
...
     { label: messages.operatorDropdownIsNotOptionText, value: '::!' },
     { label: messages.operatorDropdownContainsOptionText, value: ':' },
     { label: messages.operatorDropdownDoesNotContainOptionText, value: ':!' }
   ];
+  const showRemoveRuleRowButton = ruleRows.rows.length > 1;
+
+  const handleRemoveRuleRowOnClick = () => {
+    setRuleRows(
+      Object.assign({}, ruleRows, {
+        rows: ruleRows.rows.filter(ruleRow => ruleRow.id !== rowId)
```
maniax89: is the plan for the `ruleRows` object to have more attributes? or could this just be an array that we do the same operation on?

also `Object.assign` can usually be replaced with `{...ruleRows, rows: ruleRows.rows.filter()}`

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -1,44 +1,81 @@
...
+        id={`structured-query-rules-value-` + rowId}
         labelText={messages.valueInputLabelText}
         placeholder={messages.valueInputPlaceholderText}
       />
+      {showRemoveRuleRowButton && (
+        <Button
+          hasIconOnly
+          kind="ghost"
+          renderIcon={SubtractAlt16}
+          iconDescription="Remove row"
```
maniax89: i18n

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
kabeaty: Yeah, my plan is that it will have more attributes. And will replace 👍

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
kabeaty: oop, good catch, thanks, will update!

PR #28, opened 2019-12-19
fix: focus styles for autocomplete term items

kabeaty:
This makes it so that the focus state for autocompletion items is no longer partially hidden by the dropdown.

related to issue: https://github.ibm.com/watson-discovery/disco-widgets/issues/443

You can test this out in Storybook with the default `SearchInput` story. Get autocompletions to show and then tab through them, seeing that the focus state shows a full blue border around each autocompletion item now.

focus states should now look like this: 
Screen Shot 2019-12-19 at 2 26 19 PM
As opposed to this:
1c3fbf80-15c2-11ea-811c-2a49825f28c5

n/a

no


kabeaty:
Yes, np, I added screenshots 👍

PR #26, opened 2019-12-19
test: cypress search input

noah-eigenfeld:
#### What do these changes do/fix?
Adds cypress tests that cover functionality of the SearchInput component. We decided to break up cypress tests by use case, as opposed to component-by-component.

Tested functionality:
- Basic search query
- Spelling correction
- Autocomplete

#### How do you test/verify these changes?
Run the cypress tests:
```
yarn workspace discovery-search-app test:e2e
```

OR use cypress's UI to run individual tests:
```
yarn workspace discovery-search-app cypress
```
#### Have you documented your changes (if necessary)?
I think we've already got enough documentation on using cypress with our project, but I can add more if someone thinks that's necessary.
#### Are there any breaking changes included in this pull request?
Nope

PR #18, opened 2019-12-09
Add Cypress tests to cover our components' functionality

noah-eigenfeld:

PR #39, opened 2020-01-02
test: search results cypress tests

noah-eigenfeld:
Adding Cypress tests for the functionality surrounding search results. Functionality covered in these tests:
- [x] Pagination
- [x] Table results
- [x] Passage results

This PR builds off of work that started with #26 and #18.

Run Cypress tests with the Cypress UI:
```
yarn workspace discovery-search-app cypress
```
OR:
Run Cypress tests without the Cypress UI:
```
yarn workspace discovery-search-app test:e2e
```
Either way, you should be able to verify that the new tests work and that they make sense for the functionality.
No.
Nope.


examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+      cy.get('.bx--pagination__text').should('contain', '12');
+    });
+
+    it('the previous page button is disabled', () => {
+      cy.get('.bx--pagination__button--backward').should('be.disabled');
+    });
+
+    describe('and the next page arrow is clicked', () => {
+      beforeEach(() => {
+        cy.wait('@multiPageQuery');
```
maniax89: this seems strange that the wait is before the click -> shouldn't it come after?

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+        });
+      });
+    });
+
+    describe('and we use the page selector', () => {
+      beforeEach(() => {
+        cy.wait('@multiPageQuery');
+      });
+
+      it('we can navigate to each page of results', () => {
```
maniax89: i think this might be a little overkill for an e2e test - i think just checking one of them should be sufficient

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+      describe('and we increase the number of results per page to 50', () => {
+        beforeEach(() => {
+          cy.get('.bx--pagination__left')
+            .find('.bx--select-input')
+            .select('50');
+        });
+
+        it('returns to the first page, with the correct size', () => {
+          cy.wait('@multiPageQuery');
+          cy.wait('@multiPageQuery').then(xhr => {
```
maniax89: is there a reason it sends 2 requests?

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    describe('and items per page is set to 5', () => {
+      beforeEach(() => {
+        cy.get('.bx--pagination__left')
+          .find('.bx--select-input')
+          .select('5');
+      });
+
+      it('makes a request for 5 results', () => {
+        cy.wait('@multiPageQuery'); // not the request we're looking for
```
maniax89: curious why we have multiple requests - shouldn't it only be submitting one `POST /query`?

examples/discovery-search-app/cypress/integration/pagination/spec.ts
ChrisGillin: If this is moved after the button click, can it also be removed from all of the following tests?

ChrisGillin: Is this waiting on the query from https://github.com/watson-developer-cloud/discovery-components/pull/39/files#diff-5cfb6b041f06b8eb649bc618aa84fbb2R46 and the waits waiting for the request from the click below? If so, doesn't like 45 handle making sure the initial request went through? It looks like a lot of the individual tests start off with a `wait`, is there a way to restructure them to move all those `wait`s into `beforeEach`s

examples/discovery-search-app/cypress/integration/pagination/spec.ts
ChrisGillin: Is one of these to retrieve the facets and one for the actual result set and if so are we unable to skip that initial query if we aren't requesting facets?

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+      });
+
+      it('only lists twelve pages of results', () => {
+        cy.wait('@multiPageQuery');
+        cy.wait('@multiPageQuery');
+        cy.get('.bx--pagination__text').should('contain', '12');
+      });
+    });
+
+    describe('and items per page is set to 10', () => {
```
ChrisGillin: Isn't this already covered by the previous `describe` and by checking the number of pages with the default page size?

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+      });
+
+      it('only lists six pages of results', () => {
+        cy.wait('@multiPageQuery');
+        cy.wait('@multiPageQuery');
+        cy.get('.bx--pagination__text').should('contain', '6');
+      });
+    });
+
+    describe('and items per page is set to 50', () => {
```
ChrisGillin: Ditto

examples/discovery-search-app/cypress/integration/basic/spec.ts
```diff
@@ -1,6 +1,6 @@
 describe('Basic search', () => {
   beforeEach(() => {
-    cy.server({ force404: true });
```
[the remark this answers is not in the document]
noah-eigenfeld: Removed it, because it's getting set automatically in the `serverDefaults` file: https://github.com/watson-developer-cloud/discovery-components/blob/d9b633b51706604102d8c08541a96b56f2aac28d/examples/discovery-search-app/cypress/support/serverDefaults.ts#L2

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+  beforeEach(() => {
+    cy.server();
+    cy.fixture('query/query.json').as('queryJSON');
+    cy.fixture('collections/collections.json').as('collectionsJSON');
+    cy.fixture('component_settings/componentSettings.json').as('componentSettingsJSON');
+    cy.fixture('query/multiPageResults.json').as('multiPageResultsJSON');
+    cy.route('GET', '**/component_settings?version=2019-01-01', '@componentSettingsJSON').as(
+      'componentSettings'
+    );
+    cy.route('POST', '**/query?version=2019-01-01', '@queryJSON').as('searchQuery');
```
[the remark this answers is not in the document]
noah-eigenfeld: Definitely a good call. For some of the specific cases, we could end the name with the thing that makes the object unique, such as:
- `postQueryNoResults`
- `postQueryPassageResults`
- `postQueryMisspelled`
and so on

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    cy.route('GET', '**/collections?version=2019-01-01', '@collectionsJSON').as('getCollections');
+    cy.visit('/');
```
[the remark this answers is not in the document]
noah-eigenfeld: I haven't had a problem with that so far, but I guess it couldn't hurt to add

examples/discovery-search-app/cypress/integration/pagination/spec.ts
noah-eigenfeld: I remember digging into the multiple query requests with @Steven-Gassert before the holidays. I think we narrowed it down to the facets, just as Gillin is describing. I'm not sure if there's a way to stop the app from sending the facets request, but we could probably make this a little bit cleaner by doing
```
cy.wait([@multiPageQuery, @multiPageQuery])
  .spread((_, multiPageQuery) => {
    //evaluate @multiPageQuery
  });
```

maniax89: seems like a bug if we have multiple requests being sent at the same time. i think we have a request being sent at page load, but we shouldn't be having multiple for pagination clicks

ChrisGillin: I missed the fact there were multiple _per click_. Yeah that sounds bad.

examples/discovery-search-app/cypress/integration/pagination/spec.ts
noah-eigenfeld: That `wait` is to make sure the query went through from the `describe` this is a child of. When writing these tests, I put the wait into either the `it`s or the following `describe`s, because I usually had one `it` check that all of the request params were correct. Looking at this now, though, I bet I could move the wait up by doing something like:
```
describe('click a button' () => {
  beforeEach(() => {
    button.click();
    cy.wait('@request').as('clickedRequest');
  });

  it('the request has the correct params', () => {
    cy.get('@clickedRequest').its('whatever.param').eq('correct.value');
  )};
});
```

noah-eigenfeld: I didn't think to try aliasing the request caught by `wait` before... I'll give it a shot and see how it goes.

noah-eigenfeld: It worked! 🎆

examples/discovery-search-app/cypress/integration/pagination/spec.ts
noah-eigenfeld: I cleaned up a bunch of the `cy.wait`s, and I think this is just the one page-load request, not a bunch of duplicates on subsequent clicks. Handling the `wait`s in the `beforeEach`s cut down on the sheer number, and adding one to the very beginning, right after `cy.visit('/')` removed the need for any of the other double-`wait`s. I'll push up an update that shows these changes shortly...

examples/discovery-search-app/cypress/integration/basic/spec.ts
```diff
@@ -1,12 +1,21 @@
...
+    //TODO: move this section of code into support.js? Maybe just as a helper function
```
noah-eigenfeld: Looking for feedback on this. This is a large chunk of code that just handles getting the server setup, navigating us to the example app, and mocking out the requests that happen on page load. I could put this into a `beforeEach` in `cypress/support/index.js`, where it would run before *every* test file. That might make it a little less apparent to future developers if they need to override any of these initial routes. Maybe it would be better to define a `loadHomePage()` helper method that does all of this? That would make it easier to see "these are the things we do by default, but you're welcome to do something different."

examples/discovery-search-app/cypress/integration/pagination/spec.ts
noah-eigenfeld: Not sure if this comment is still relevant after the last couple commits I put in. I didn't want to manually test all of the options in the Items per page dropdown, so I just picked 5 (the default size), 10, and 50, to get a good spread.

examples/discovery-search-app/cypress/integration/basic/spec.ts
noah-eigenfeld: In order to make this work, I had to get typescript working in our cypress tests. I think it might be best to keep those changes isolated from this PR, so I put them in #44, and we can merge it into `develop` after this PR goes in.

PR #42, opened 2020-01-06
feat: structured query add and remove group rules

kabeaty:
These changes include the ability to add and remove groups of rules for the `StructuredQuery` component. I set the default max number of rule groups that can be added beneath the top-level to 5 for now.

You can manually test out and play around with these changes by adding and removing group rules and their rule rows in the `StructuredQuery` Storybook. I've also added tests for adding/removing rule groups, and for adding/removing rule rows from these rule groups.

n/a

no


packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+   */
+  setStructuredQuerySelection: Dispatch<SetStateAction<StructuredQuerySelection>>;
+}
+
+const handleOnClick = (
+  structuredQuerySelection: StructuredQuerySelection,
+  setStructuredQuerySelection: Dispatch<SetStateAction<StructuredQuerySelection>>,
+  groupId: number | 'top-level'
+) => {
+  if (groupId === 'top-level') {
```
ChrisGillin: Would it cut out any additional code to just check if it's the first group (or id == 0), rather than making the id `number | 'top-level'`, or is it not guaranteed to be first (or have 0 as the id)?

kabeaty: Yeah, the top-level group doesn't have an `id` at all as I have it set up, since there can only be one top-level that contains rows and groups. Which is why I ended up using the `top-level` here. If it had an `id`, then would need another way to know that it was top-level and not nested.

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
```diff
@@ -19,41 +19,50 @@ export interface StructuredQueryProps {
...
       <div className={structuredQueryRulesButtonsClass}>
         {showAddRuleRowButton && (
-          <Button kind="ghost" renderIcon={Add16} onClick={handleAddRuleRowOnClick}>
-            {mergedMessages.addRuleRowText}
-          </Button>
+          <AddRuleRowButton
+            addRuleRowText={mergedMessages.addRuleRowText}
+            setStructuredQuerySelection={setStructuredQuerySelection}
+            structuredQuerySelection={structuredQuerySelection}
+            groupId="top-level"
```
maniax89: why not just `0`?

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export interface AddRuleRowButtonProps {
+  /**
+   * text to display for the Add rule button
+   */
+  addRuleRowText: Messages['addRuleRowText'];
+  /**
+   * id of the group for the rule row to render, or 'top-level' if the top-level rule group
+   */
+  groupId: number | 'top-level';
```
maniax89: this feels strange - can we store something else like `0` to indicate the root level?

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+   * state that represents the current rules and selections for the structured query
+   */
+  structuredQuerySelection: StructuredQuerySelection;
+  /**
+   * used to set the structuredQuerySelection state
+   */
+  setStructuredQuerySelection: Dispatch<SetStateAction<StructuredQuerySelection>>;
+}
+
+const handleOnClick = (
```
maniax89: why did you move this outside of the component?

packages/discovery-react-components/src/components/StructuredQuery/components/RemoveRuleRowButton/RemoveRuleRowButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const handleOnClick = () => {
+    if (groupId === 'top-level') {
+      setStructuredQuerySelection({
+        ...structuredQuerySelection,
+        rows: structuredQuerySelection.rows.filter((ruleRow: Row) => ruleRow.id !== rowId)
+      });
+    } else {
+      setStructuredQuerySelection({
+        ...structuredQuerySelection,
```
maniax89: something to be careful of here - this is only doing a shallow copy. i don't think there are any gotchas with this code as it stands, but could turn hard-to-debug issues later on (nested objects being mutated)

packages/discovery-react-components/src/components/StructuredQuery/components/RuleGroup/RuleGroup.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const isTopLevelGroup = groupId === 'top-level';
+  let rows: Row[] = structuredQuerySelection.rows;
+  if (!isTopLevelGroup) {
+    structuredQuerySelection.groups.map(group => {
+      if (group.id === groupId) {
+        rows = group.rows;
+      }
+    });
+  }
+  const showAddRuleRowButton = rows.length < MAX_NUM_SIBLING_RULE_ROWS && !isTopLevelGroup;
```
maniax89: do we not want to allow a rule addition at the top level?

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleGroupButton/AddRuleGroupButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const AddRuleGroupButton: FC<AddRuleGroupButtonProps> = ({
+  addRuleGroupText,
+  structuredQuerySelection,
+  setStructuredQuerySelection
+}) => {
+  const handleOnClick = () => {
+    const newRuleGroupId: Group['id'] =
+      structuredQuerySelection.groups.length !== 0
+        ? structuredQuerySelection.groups[structuredQuerySelection.groups.length - 1].id! + 1
```
maniax89: why does this need a `!`?

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
kabeaty: I refactored with a data structure similar to what you recommended above (just with the `entities` part, except I'm calling it `groups`), so this is now `0` for top-level groups.

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
kabeaty: Done - see previous note - `0` now represents top level

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
kabeaty: I think this happened during splitting this out into its own component. Thanks for catching, updated!

packages/discovery-react-components/src/components/StructuredQuery/components/RuleGroup/RuleGroup.tsx
kabeaty: I'm checking for this here since we already include the `Add rule` button for the top-level in `StructuredQuery` to make its `Add rule` button appear all the way at the bottom. I could add a comment if you think that would help clarify in future?

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleGroupButton/AddRuleGroupButton.tsx
kabeaty: With refactoring, this is no longer here

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
```diff
@@ -35,9 +35,9 @@ export const AddRuleRowButton: FC<AddRuleRowButtonProps> = ({
     groupId: number
   ) => {
     const newRuleRowId =
-      structuredQuerySelection.groups[groupId].rows[
-        structuredQuerySelection.groups[groupId].rows.length - 1
-      ] + 1;
+      structuredQuerySelection.groups[groupId].rows.reduce(function(a, b) {
```
maniax89: could just do `reduce((a, b) => Math.max(a, b))` -> also would prefer more descriptive variable names if possible

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
```diff
@@ -19,41 +20,49 @@ export interface StructuredQueryProps {
 
 const StructuredQuery: FC<StructuredQueryProps> = ({ messages = defaultMessages }) => {
   const mergedMessages = { ...defaultMessages, ...messages };
-  const [ruleRows, setRuleRows] = useState({ rows: [{ id: 0 }] });
-  const showAddRuleRowButton = ruleRows.rows.length < MAX_NUM_SIBLING_RULE_ROWS;
+  const [structuredQuerySelection, setStructuredQuerySelection] = useState<
```
maniax89: quick question - why is it `structuredQuerySelection` vs just `structuredQuery`? just curious what the `selection` part is for

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
```diff
@@ -19,41 +20,49 @@ export interface StructuredQueryProps {
...
-    const newRuleRow = { id: newRuleRowId };
-    setRuleRows({
-      ...ruleRows,
-      rows: ruleRows.rows.concat(newRuleRow)
-    });
-  };
+  const showAddRuleRowButton =
+    structuredQuerySelection.groups[0].rows.length < MAX_NUM_SIBLING_RULE_ROWS;
+  const showAddRuleGroupButton =
+    keys(structuredQuerySelection.groups).length - 1 < MAX_NUM_NESTED_RULE_GROUPS;
```
maniax89: is there a difference between `keys` and `Object.keys`?

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleGroupButton/AddRuleGroupButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const AddRuleGroupButton: FC<AddRuleGroupButtonProps> = ({
+  addRuleGroupText,
+  structuredQuerySelection,
+  setStructuredQuerySelection
+}) => {
+  const handleOnClick = () => {
+    const maxGroupId = keys(structuredQuerySelection.groups).reduce(function(a, b) {
+      return Math.max(parseInt(a), parseInt(b)).toString();
```
maniax89: does this need to be `toString()`-ed?

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const AddRuleRowButton: FC<AddRuleRowButtonProps> = ({
+  addRuleRowText,
+  groupId,
+  structuredQuerySelection,
+  setStructuredQuerySelection
+}) => {
+  const handleOnClick = (
+    structuredQuerySelection: StructuredQuerySelection,
```
maniax89: i'm not sure you need to pass all these attributes into the function, you can just use them from the props above

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
kabeaty: Done! Thanks

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+        }
+      }
+    });
+  };
+
+  return (
+    <Button
+      kind="ghost"
+      renderIcon={Add16}
+      onClick={() => handleOnClick(structuredQuerySelection, setStructuredQuerySelection, groupId)}
```
maniax89: doesn't need to be an arrow function if you don't pass the props in (see above)

packages/discovery-react-components/src/components/StructuredQuery/components/RuleGroup/RuleGroup.tsx
maniax89: i think its fine. i'm guessing that if a rule group only has a single rule we don't wrap parentheses around the query expression that we end up generating

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
```diff
@@ -19,41 +20,49 @@ export interface StructuredQueryProps {
...
 
   return (
     <div className={structuredQueryClass}>
-      <RuleGroupDropdown messages={mergedMessages} />
-      {ruleRows.rows.map(row => {
+      {keys(structuredQuerySelection.groups).map(id => {
```
maniax89: i'm not sure this will ensure the order of the groups unless lodash's `keys` operates differently than `Object.keys`. i was thinking that there would be a second array that stores the group ids so that we know the order of them like
```
{
  "groups": {
     "0": {
       "rows": [1]
     },
     "1": {
     }
   }
   "group_order": [0]
}
```

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
kabeaty: My thinking was that it includes (and will include more) of the selections that people are making (right now, adding the rule groups, but in the future, to the operators, etc.) when building their queries. I could also be fine with just `structuredQuery`.

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleGroupButton/AddRuleGroupButton.tsx
kabeaty: Yes, otherwise TS complains because `string` is the return type expected here...

packages/discovery-react-components/src/components/StructuredQuery/components/AddRuleRowButton/AddRuleRowButton.tsx
kabeaty: updated - this was related to moving this `handleOnClick` outside of the component when I split it out - thanks!

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
kabeaty: Ah, ok, I see what you mean here, good point. Updated with a `group_order` 👍

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
kabeaty: I thought there was a reason I used that instead, but actually I think there isn't a difference, so updated to just use `Object.keys` 👍

PR #43, opened 2020-01-08
fix: offest issue in CIDocument

broulaye:
This changes fixes the highlighting of the attributes in CIDocument, you can see it's working properly by running storybook `yarn storybook` and navigating to `CIDocument -> default`. select the highlights document and select defined terms under attributes you should see that  everything is highlighted correctly.

Yes
No


anibapat:
This fixes the problem with attributes, however other highlights gets messed up for eg. 

Before: 
Screen Shot 2020-01-08 at 10 07 07 AM

 
After this change: 
Screen Shot 2020-01-08 at 10 05 17 AM



In invoices: 

Before:  
Screen Shot 2020-01-08 at 10 07 26 AM


After: 
Screen Shot 2020-01-08 at 10 05 06 AM

broulaye:
@anibapat this is ready for a re-review.

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -250,6 +251,14 @@ function setupSectionParser(
     },
 
     ontext: (_: Parser, text: string): void => {
+      if (decodeHTML(text).length != text.length) {
```
jhpedemonte: Add a comment here: why are we doing this block?

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -250,6 +251,14 @@ function setupSectionParser(
...
+        const lastElem = sectionHtml.length - 1;
```
jhpedemonte: I think this is wrong. Since you're making changes to the _open tag_, do same as in `onclosetag()` below: get the index from `openTagIndices`.

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -250,6 +251,14 @@ function setupSectionParser(
...
+        sectionHtml[lastElem] = sectionHtml[lastElem].replace(
+          />$/,
+          ` original-text="${encodeHTML(text)}">`
```
jhpedemonte: Add a comment as to why we're encoding the text here.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +88,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
     throw new Error(`Failed to find text node. Node: ${node.textContent}, offset: ${offset}`);
   }
 
-  const textOffset = Math.max(0, offset - runningOffset);
+  let textOffset = Math.max(0, offset - runningOffset);
+
+  const originalText = nodeElement.getAttribute('original-text');
```
jhpedemonte: You can move this to right before the `if` below, without empty lines.

packages/discovery-react-components/src/utils/document/processDoc.ts
jhpedemonte: Also, make this a `data` attribute: `data-orig-text`

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +88,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
...
+
+  // If attribute 'original-text' is present then
+  // the string contains some encoded entities
+  // so, we need adjust the text offset
+  if (originalText) {
+    // To properly calculate the offset of the string we want to highlight
+    // we need to get the offset from the decoded html text and subtract
+    // it from the original encoded html offset
```
jhpedemonte: Update this comment to specify that the offsets are based on the HTML string, which may contain HTML entities; but need to calculate the offset in the DOM, where the HTML entities have been rendered, as single characters now. Maybe throw in an example of `&amp;` in the HTML string vs `&` in the DOM.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +88,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
...
+    const encodedTextSubstring = originalText.substring(0, textOffset);
+
+    const decodedText = decodeHTML(encodedTextSubstring);
+
+    const adjustment = textOffset - decodedText.length;
+
+    textOffset = Math.max(0, textOffset - adjustment);
```
jhpedemonte: I think you can remove all the empty lines above.

jhpedemonte:
Functionally, it works great with our sample docs. @anibapat please also take a look, since you found issues earlier.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -65,17 +66,20 @@ export function findOffsetInDOM(
 }
 
 export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
-  const nodeOffset = parseInt((node as HTMLElement).dataset.childBegin || '0', 10);
+  const nodeElement = node as HTMLElement;
+  const nodeOffset = parseInt(nodeElement.dataset.childBegin || '0', 10);
```
anibapat: Why this change is introduced? I think it can be as is previously.

broulaye: because i'm using this node element in many places below, instead of doing this `node as HTMLElement` every time, i think it's best to do it this way

anibapat: How about doing it here on line 68
```
export function getTextNodeAndOffset(node: HTMLElement, 
```

broulaye: because you don't always need it as an HTMLElement like when you pass it to the iterator.

broulaye: I don't really see the downside of having this variable.

anibapat: Oh, haven't seen that `node` is used in the iterator.  I am fine with the new variable then.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +88,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
     throw new Error(`Failed to find text node. Node: ${node.textContent}, offset: ${offset}`);
   }
 
-  const textOffset = Math.max(0, offset - runningOffset);
+  let textOffset = Math.max(0, offset - runningOffset);
+
+  // If attribute 'original-text' is present then
```
jhpedemonte: -> `data-orig-text`

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +88,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
...
+  // so, we need adjust the text offset
+  const originalText = nodeElement.getAttribute('data-orig-text');
+  if (originalText) {
+    // To properly calculate the offset of the string we want to highlight
+    // we need to get the offset from the decoded html text and subtract
+    // it from the original encoded html offset. we doing this because the
+    // offsets are based on the original html string which may contain
+    // html entities (eg. &quot;), however those entities get rendered in
+    // the dom which throws off the offsets values, for example &quot;
+    // becomes ", going from 6 characters to 1.
```
jhpedemonte: Wrap `&quot;` and `"` in backticks (`).

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +88,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
     throw new Error(`Failed to find text node. Node: ${node.textContent}, offset: ${offset}`);
   }
 
-  const textOffset = Math.max(0, offset - runningOffset);
+  let textOffset = Math.max(0, offset - runningOffset);
+
+  // If attribute 'original-text' is present then
+  // the string contains some encoded entities
+  // so, we need adjust the text offset
+  const originalText = nodeElement.getAttribute('data-orig-text');
```
jhpedemonte: For consistency with the other code, you could reference this as `nodeElement.dataset.origText` (see `dataset.childBegin` above).

packages/discovery-react-components/src/utils/dom.ts
```diff
@@ -7,13 +7,12 @@ export function clearNodeChildren(node: Node): void {
...
+ * Return HTML encoded text node. Need to encode the string before measuring
+ * the length, because the offsets are calculated from the HTML string,
+ * which has encoded entities.
  *
  * @param textNode DOM Text node
  */
-export function getEncodedTextNodeLength(textNode: Text): number {
-  return encodeHTML(textNode.data).length;
+export function getEncodedTextNode(textNode: Text): string {
+  return encodeHTML(textNode.data);
```
jhpedemonte: Since this is just a one-liner and only used in a single place, remove it and use `encodeHTML(textNode.data)` directly.

jhpedemonte:
Also, rebase this to the `january_fix_pack` branch.

broulaye:
closing this in favor of this PR: https://github.com/watson-developer-cloud/discovery-components/pull/46

PR #44, opened 2020-01-08
chore: Cypress + typescript

noah-eigenfeld:
To make imports and exports work in our cypress tests, we had to either switch them all back to `.js` files or make typescript work with our cypress setup.

There's a lot of cleaning up to do, but at the time of this PR's creation, these changes got Typescript working and all of the tests passing.

NOTE: this PR currently branches off of #39, because both of these PRs make significant changes to our cypress code. Hopefully that will make this PR easier to merge down the line as well.

No, but I should.


examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -70,6 +61,7 @@ describe('Pagination', () => {
 
       it('correctly requests the next page', () => {
         cy.get('@nextPageQueryObject')
+          // @ts-ignore
```
noah-eigenfeld: Cypress's typing for `its()` only allows a limited selection of strings to be passed in, even though any valid property string will work. It looks like a proper fix might involve defining our own typing for `its`, but I didn't research too far down that rabbit hole.

examples/discovery-search-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
noah-eigenfeld: This file was created by https://github.com/bahmutov/add-typescript-to-cypress, but it seems like some of these settings are covered in `cypress/tsconfig.json`. I'm guessing we could kill one of those files, and just hang on to the superset of configs.

package.json
```diff
@@ -39,7 +40,7 @@
     "carbon-components-react": "^7.7.0",
     "carbon-icons": "^7.0.7",
     "cross-env": "^5.1.4",
-    "cypress": "3.2.0",
+    "cypress": "3.8.1",
```
noah-eigenfeld: Note: when I updated this setting, then re-ran `yarn`, it automatically downloaded a new Cypress binary (runs the Cypress UI), which took like 40 minutes. Subsequent `yarn` runs didn't download this extra piece, so it should be a one-time thing for devs. Just something to be aware of when checking out this branch.

examples/discovery-search-app/package.json
```diff
@@ -33,7 +35,8 @@
     "ibm-watson": "5.2.0-alpha-1",
     "prop-types": "^15.6.2",
     "react": "^16.4.1",
-    "react-dom": "^16.4.1"
+    "react-dom": "^16.4.1",
+    "ts-loader": "^6.2.1"
```
noah-eigenfeld: I wonder if I can move these to dev dependencies at the root directory?

noah-eigenfeld:
Planning to remove the `.DS_Store` files in the next commit

examples/discovery-search-app/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
noah-eigenfeld: Note to self: do we still need this? I don't remember why I created it, and it's located in a weird spot relative to the rest of these changes.

examples/discovery-search-app/cypress/integration/pagination/spec.ts
maniax89: i'd be interested to investigate this more - `// @ts-ignore` will always throw a red flag for me (does this only apply to the following line?)

that being said - maybe if we leave a TODO comment or something similar to revisit this ill be ok with it for now. right now future developers might not know why this is here

maniax89: also where is the list of limited strings that can be passed in? if it is just `requestBody` maybe we might be using this in a way that cypress doesn't commit to supporting in future versions

examples/discovery-search-app/cypress/plugins/cy-ts-preprocessor.js
```diff
@@ -0,0 +1,26 @@
+const wp = require('@cypress/webpack-preprocessor');
+
+const webpackOptions = {
+  resolve: {
+    extensions: ['.ts', '.js']
```
maniax89: did we change everything to `.ts`? if so, why do we need the `.js` extension?

examples/discovery-search-app/cypress/support/commands.ts
```diff
@@ -1,3 +1,5 @@
+export {}; //forces this file to be considered a module, which prevents a TS compilation error
```
maniax89: nitpick - would this be better at the bottom?

examples/discovery-search-app/cypress/support/utils/defaultHomepage.ts
```diff
@@ -0,0 +1,16 @@
+//TODO: break down the logic in this funtion so that these defaults can be overwritten. We'll probably want to
+// create more util files to make this functionality more intuitive and modular.
+export function mockDefaultHomePage() {
```
maniax89: i'm guessing we went with the utils route over a custom function route so that we might structure this into classes in the future? i'm fine with that approach as it makes it clear what functions are cypress and which ones are our abstractions

examples/discovery-search-app/cypress/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "baseUrl": "../node_modules",
     "target": "es5",
     "lib": ["es5", "dom"],
-    "types": ["cypress"]
+    "types": ["cypress", "jest"]
```
maniax89: what does `jest` have to do with cypress?

examples/discovery-search-app/package.json
```diff
@@ -21,8 +21,10 @@
   },
   "dependencies": {
     "@carbon/icons": "^10.5.0",
+    "@cypress/webpack-preprocessor": "^4.1.1",
     "@ibm-watson/discovery-react-components": "^1.1.0-beta.10",
     "@ibm-watson/discovery-styles": "^1.1.0-beta.5",
+    "@types/jest": "^24.0.25",
```
maniax89: what's this for? also i think if anything this should be a dev dependency

examples/discovery-search-app/package.json
maniax89: this should be a dev dependency as it isn't required to run the example application

examples/discovery-search-app/src/react-app-env.d.ts
maniax89: i would try removing it and seeing what happens. if it complains, i think leaving it in is fine. i doubt you need to make any changes to anything under `/discovery-search-app/src` for cypress testing, though

examples/discovery-search-app/tsconfig.json
maniax89: yeah i think it makes sense to move this into the cypress directory. im worried leaving it in `discovery-search-app` will cause confusion that the search app is written in typescript, which it is not

examples/discovery-search-app/cypress/integration/pagination/spec.ts
noah-eigenfeld: Found the type definition for `its()`, but I haven't tracked down a full list of strings yet.

maniax89: yeah - it looks like the following properties are the only ones that would be permitted by that definition https://github.com/cypress-io/cypress/blob/develop/cli/types/index.d.ts#L4467
```
  interface WaitXHR {
    duration: number
    id: string
    method: HttpMethod
    request: {
      body: string | ObjectLike
      headers: ObjectLike
    }
    requestBody: WaitXHR['request']['body']
    requestHeaders: WaitXHR['request']['headers']
    response: {
      body: string | ObjectLike
      headers: ObjectLike
    }
    responseBody: WaitXHR['response']['body']
    responseHeaders: WaitXHR['response']['headers']
    status: number
    statusMessage: string
    url: string
    xhr: XMLHttpRequest
  }
```

so i imagine drilling down into anything under `responseBody` wouldn't be allowed

examples/discovery-search-app/cypress/plugins/cy-ts-preprocessor.js
noah-eigenfeld: We don't 😄

noah-eigenfeld: removing it now

examples/discovery-search-app/cypress/support/commands.ts
noah-eigenfeld: Doesn't matter to me. If #45 gets merged before this PR, then I don't think we'll even need this line.

examples/discovery-search-app/cypress/support/utils/defaultHomepage.ts
noah-eigenfeld: Yeah, that was my goal when setting this up. I don't think we need as much scaffolding as Tooling has, necessarily, but we could definitely cut down on repeated code by setting up default mocks for certain components.

examples/discovery-search-app/cypress/support/utils/defaultHomepage.ts
```diff
@@ -0,0 +1,16 @@
...
+  cy.fixture('collections/collections.json').as('collectionsJSON');
+  cy.route('GET', '**/collections?version=2019-01-01', '@collectionsJSON').as('getCollections');
+  cy.fixture('component_settings/componentSettings.json').as('componentSettingsJSON');
+  cy.route('GET', '**/component_settings?version=2019-01-01', '@componentSettingsJSON').as(
+    'getComponentSettings'
+  );
+  cy.fixture('query/query.json').as('queryJSON');
+  cy.route('POST', '**/query?version=2019-01-01', '@queryJSON').as('postQuery');
+  cy.visit('/');
+  cy.wait(['@getCollections', '@getComponentSettings', '@postQuery']);
```
noah-eigenfeld: Note to self: make it possible to pass in overrides for these fixtures. I'm thinking specifically of the facets tests, where we want to change the `componentSettings` response that comes in on page load.

examples/discovery-search-app/cypress/tsconfig.json
noah-eigenfeld: I think I added this to get rid of an error message about missing types for jest assertions within the cypress tests (such as in autocomplete/spec.ts). But I removed this now, and nothing has broken so far... Maybe whatever was throwing the error also got cleaned out in my initial pruning.

examples/discovery-search-app/cypress/integration/pagination/spec.ts
noah-eigenfeld: I'm doing some more digging on this right now. Looks like this problem is only exacerbated when the `wait`ed request gets aliased and fetched later. `cy.get('@alias')` has a return type of `JQuery<HTMLElement>`, which doesn't even have the `requestBody` parameter on it.

Here's an article with a possible workaround, but I still don't think it would let us drill father down than `requestBody`: https://medium.com/@NicholasBoll/cypress-io-making-aliases-type-safe-b6f5db07d580

noah-eigenfeld: Here's another potential solution, but it would require us to define Interfaces for the requests that we want to send. Personally, I like the previous solution better, but I wanted to record it as an option:
https://github.com/cypress-io/cypress/issues/3068

.gitignore
```diff
@@ -7,4 +7,6 @@ coverage
 .npmrc
 
 lerna-debug.log
-yarn-error.log
\ No newline at end of file
+yarn-error.log
+
+**/.DS_Store
```
noah-eigenfeld: This change isn't connected to the typescript stuff; I've just seen these files popping up locally, and figured we should add them to the gitignore

package.json
```diff
@@ -73,6 +73,7 @@
     "source-map-explorer": "2.0.1",
     "start-server-and-test": "^1.10.0",
     "tsconfig-paths-webpack-plugin": "^3.2.0",
+    "ts-loader": "^6.2.1",
```
maniax89: so we have `awesome-typescript-loader` and `ts-loader` -> not a big deal, but im curious why we need both

noah-eigenfeld: Looks like we're using `awesome-typescript-loader` over in our packages, and `ts-loader` is the one that got added in our example app.. Interesting... I think it's definitely worth doing some compare/contrast and deciding on just one to use for both. It would feel weird if TS worked slightly differently in our tests vs our components.

`ts-loader` is just what I saw from other examples. I'm curious if there was a reason we went with `awesome-typescript-loader` for our components?

maniax89: i think it was just faster for storybook

PR #45, opened 2020-01-08
test: Add Cypress Tests for Facets

noah-eigenfeld:
Adds cypress tests against the following pieces of functionality:
- [x] Filter by collection
- [x] Single-select facets
- [x] Multi-select facets
- [x] Dynamic facets

Once this PR is merged in, the major pieces of functionality in our components should be covered by cypress tests.

This PR builds off work from #39 and #44.

Run Cypress tests with the Cypress UI:
```
yarn workspace discovery-search-app cypress
```
OR:
Run Cypress tests without the Cypress UI:
```
yarn workspace discovery-search-app test:e2e
```
Either way, you should be able to verify that the new tests work and that they make sense for the functionality.
No
Shouldn't be


examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
+describe('Collection Filter', () => {
+  beforeEach(() => {
+    //TODO: swap this section with the helper function, once that code is merged in
```
noah-eigenfeld: Once #44 gets merged, this block of code can be replaced with a helper method

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    cy.wait(['@getCollections', '@getComponentSettings', '@postQuery']);
+
+    // Set up/override routes & fixtures that are specific to this file
+    cy.fixture('query/singleCollectionQuery.json').as('singleCollectionQueryJSON');
+    cy.fixture('query/doubleCollectionQuery.json').as('doubleCollectionQueryJSON');
+  });
+
+  describe('when the example app loads', () => {
+    it('the collection facet select does not appear', () => {
+      cy.get('.collection-facet-select').should('not.exist');
```
maniax89: i would be careful about doing a `not.exist` check prior to at least one `exist` call - this would pass probably immediately without something else to look for on the page

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        });
+      });
+
+      describe('and we click the selection box again', () => {
+        beforeEach(() => {
+          cy.get('#collection-facet-select').click();
+        });
+
+        it('the collection filter dropdown disappears', () => {
+          cy.get('#collection-facet-select__menu').should('not.exist');
```
maniax89: i would make sure something exists in tandem with the `not.exist` to make sure we aren't asserting a white screen is a pass here

examples/discovery-search-app/cypress/integration/autocomplete/spec.ts
```diff
@@ -31,7 +31,7 @@ describe('Autocomplete', () => {
       cy.wait('@getAutocompletions');
     });
 
-    it('displays a dropdown of autocomplete suggestions', () => {
+    it.only('displays a dropdown of autocomplete suggestions', () => {
```
maniax89: guessing this is temporary?

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -24,7 +24,7 @@ describe('Passage Results', () => {
     });
 
     it('SearchResults displays ONLY the first passage text of the results that have passages', () => {
-      cy.get('.bx--search-result__content-wrapper__body')
+      cy.get('div[data-testid="search-result-element-body-passage"]')
```
maniax89: we could bring in the https://github.com/testing-library/cypress-testing-library to provide a native `getByTestId`

noah-eigenfeld:
I still need to read through the existing code comments, but since I just pushed up the last group of tests to cover, I'm moving this PR out of draft mode. Planning to dive into the reviews tomorrow.

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
...
+            .contains('finnegans wake')
+            .click();
+          cy.wait('@postQuerySingleCollection').as('singleCollectionQueryObject');
+        });
+
+        it('should make a query against only the selected collection', () => {
+          cy.get('@singleCollectionQueryObject')
+            //@ts-ignore
+            .its('requestBody.collection_ids')
+            .should('contain', 'paris19221939');
```
kabeaty: Is there a way to check that it contains only this collection_id? Below, I noticed you chained something similar with `and`, so it seems like this could technically be querying against multiple collections but we're only checking that it includes this one and not also that it doesn't include any others.

examples/discovery-search-app/cypress/integration/dynamic_facets/spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    // Sets up and handles the collections, component settings, and initial query requests that run on page-load
+    cy.server();
+    cy.fixture('collections/collections.json').as('collectionsJSON');
+    cy.route('GET', '**/collections?version=2019-01-01', '@collectionsJSON').as('getCollections');
+
+    //override the standard query with one that has more facets
+    cy.fixture('query/facetsQuery.json').as('facetsQueryJSON');
+    cy.route('POST', '**/query?version=2019-01-01', '@facetsQueryJSON').as('postQueryFacets');
+
+    //override the component settings to make facets single-select
```
kabeaty: Do you mean `multi-select` in this comment?

examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
```diff
@@ -0,0 +1,190 @@
...
+    // Sets up and handles the collections, component settings, and initial query requests that run on page-load
+    cy.server();
+    cy.fixture('collections/collections.json').as('collectionsJSON');
+    cy.route('GET', '**/collections?version=2019-01-01', '@collectionsJSON').as('getCollections');
+
+    //override the standard query with one that has more facets
+    cy.fixture('query/facetsQuery.json').as('facetsQueryJSON');
+    cy.route('POST', '**/query?version=2019-01-01', '@facetsQueryJSON').as('postQueryFacets');
+
+    //override the component settings to make facets single-select
```
kabeaty: Do you mean `multi-select` in this comment too? (btw, I find the comments helpful)

examples/discovery-search-app/cypress/integration/passages/spec.ts
kabeaty: Does it make sense to use snake_case here to align in how our `data-test-id`s are being formatted in Tooling? Just thinking of some ways we could maintain consistency to minimize the friction when switching between the projects

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
noah-eigenfeld: That's a good call. I'll add something to check for.

noah-eigenfeld: Actually, looking closer at this test, it isn't even working. `collection-facet-select` is the id of the component, not a classname. And the collection facet selector *is* visible on page load; this test was just written before I had the `wait`s for initial requests in place.

I think I'll switch this test to "the collection facet selector ~does not~ appears."

examples/discovery-search-app/cypress/integration/autocomplete/spec.ts
noah-eigenfeld: Whoops, yeah, that needs to get axed. I think I added that during the demo I gave, then forgot to remove it, because I wasn't touching the autocomplete tests. Removing...

Steven-Gassert:
PR's like this where we add a bunch of new mock data really make me interested in finding a way to type check our JSON files. Perhaps our repo would be a good place to implement something like this before pulling it into the Tooling repo?

noah-eigenfeld:
@Steven-Gassert I agree with you there, but I also think that's outside the scope of this issue. Depending on how much work it ends up being, that's maybe something that could be squeezed into #44. Not knowing the size of the task, though, I'm more inclined to make typing our fixtures its own issue.

Steven-Gassert:
Yup agreed, was just a thought as I was scanning over this PR. Spun up an issue here: https://github.ibm.com/Watson-Discovery/disco-widgets/issues/468

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  describe('when a query is made', () => {
+    beforeEach(() => {
+      cy.get('.bx--search-input').type('abil{enter}');
+      cy.wait('@postQuery');
+    });
+
+    it('the collection selector should appear', () => {
+      cy.get('#collection-facet-select').should('exist');
```
Steven-Gassert: Would cypress testing library effect the way you could select these collection facets?

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        it('should make a query against only the selected collection', () => {
+          cy.get('@singleCollectionQueryObject')
+            //@ts-ignore
+            .its('requestBody.collection_ids')
+            .should('contain', 'paris19221939');
+        });
+
+        it('the clear all selected collections button appears', () => {
+          cy.get('div[aria-label="Clear Selection"]').should('exist');
```
Steven-Gassert: I wonder if data test id's would be better here. I'm envisioning a scenario where we get a second round of design revisions and we need to make tweaks to the naming we've used for certain buttons and labels

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,142 @@
...
+          });
+        });
+      });
+
+      describe('and we click the selection box again', () => {
+        beforeEach(() => {
+          cy.get('#collection-facet-select').click();
+        });
+
+        it('the collection filter dropdown disappears', () => {
```
Steven-Gassert: Not sure if this is too verbose and we just want to trust that carbon will take care of this for us. But perhaps putting in a test here to make sure if a click even happens outside of the collection filter dropdown. that the dropdown also disappears. I only bring this up because I remember seeing bugs like this in tooling that went uncaught

examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
noah-eigenfeld: yeah -- forgot to change that 😄

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
noah-eigenfeld: Yeah, I could see this being too much detail. I don't have a strong feeling, but I'm leaning towards leaving it here. Maybe something to get rid of if we start breaking our tests down into smaller files.

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
noah-eigenfeld: It definitely could. I think we could swap this out for
```
cy.getById('collection-facet-select').should('exist');
```

Or, we could use some other parameter to find this element by.

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
Steven-Gassert: Before I was wondering if we needed a test to cover that the collections facet closes when a click event happens outside of the facet component, because I've seen this error make it's way into Tooling. Probably not important enough to add if this isn't as simple as copy and paste this test, and then put the click event else where

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
noah-eigenfeld: I could do either
```
.should('eq', ['paris19221939']);
```
or
```
.should('contain', 'paris19221939')
.and('have.length', 1);
```

The former is probably cleaner

kabeaty: Yeah, the former looks clear to me 👍

examples/discovery-search-app/cypress/integration/basic/spec.ts
```diff
@@ -21,18 +21,18 @@ describe('Basic search', () => {
   // Rendering initial page
   describe('When the example app loads', () => {
     it('SearchInput has placeholder text "Search"', () => {
-      cy.get('.bx--search-input').should('have.attr', 'placeholder', 'Search');
+      cy.findByPlaceholderText('Search').should('have.attr', 'placeholder', 'Search');
```
kabeaty: Since it's finding by the placeholder text now, isn't the `should` check essentially checking for the same thing it's found by? In this case, I think it could be clearer to just test for should exist.

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    cy.wait(['@getCollections', '@getComponentSettings', '@postQuery']);
+
+    // Set up/override routes & fixtures that are specific to this file
+    cy.fixture('query/singleCollectionQuery.json').as('singleCollectionQueryJSON');
+    cy.fixture('query/doubleCollectionQuery.json').as('doubleCollectionQueryJSON');
+  });
+
+  describe('when the example app loads', () => {
+    it('the collection facet select does not appear', () => {
+      cy.contains('Available collections').should('exist');
```
kabeaty: I might be missing something, but the `it` description makes it sound like the collection facet isn't expected to exist but then this is testing that it should exist?

examples/discovery-search-app/cypress/integration/basic/spec.ts
```diff
@@ -21,18 +21,18 @@ describe('Basic search', () => {
...
-      cy.get('.bx--search-magnifier').should('be.visible');
+      cy.get('svg.bx--search-magnifier').should('be.visible');
     });
   });
 
   // Querying with results
   describe('When entering a query with results', () => {
     beforeEach(() => {
-      cy.get('.bx--search-input').type('abil{enter}');
+      cy.findByPlaceholderText('Search').type('abil{enter}');
```
kabeaty: Does it make sense to add an alias for `@searchInput` in this file too for consistency?

noah-eigenfeld: Yeah, I'm down with that

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
noah-eigenfeld: Ah, forgot to update the description when that test changed. Fixing...

PR #46, opened 2020-01-10
fix: issue with document highlighting offset

broulaye:
This changes fixes the highlighting of the attributes in CIDocument, you can see it's working properly by running storybook

 run yarn storybook and navigating to CIDocument -> default. select defined terms under attributes you should see that everything is highlighted correctly.

Yes
Yes


packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -84,7 +85,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
     throw new Error(`Failed to find text node. Node: ${node.textContent}, offset: ${offset}`);
   }
 
-  const textOffset = Math.max(0, offset - runningOffset);
+  let textOffset = Math.max(0, offset - runningOffset);
+
+  // If attribute 'data-orig-text' is present then
```
maniax89: do we have a test for this?

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -266,6 +267,21 @@ function setupSectionParser(
     },
 
     ontext: (_: Parser, text: string): void => {
+      // In order to properly highlight texts in the DOM we need
+      // to know if the original text contains some encoded
+      // html entities, and if so, we pass that value down to
+      // be used later.
+      if (decodeHTML(text).length != text.length) {
```
maniax89: do we have a test for this?

jhpedemonte: Added a test.

packages/discovery-react-components/src/utils/document/documentUtils.ts
jhpedemonte: I took a stab, but there are some differences between browser DOM and `jsdom` (as used in tests). I'll make an issue to add some tests around this.

PR #48, opened 2020-01-13
feat: standalone HTML view with table highlighting

jhpedemonte:

Enables `HtmlView` to with table highlighting.

Screen Shot 2020-01-12 at 6 59 46 PM

https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/220
https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/221

Storybook > DocumentPreview > components > HtmlView


packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -18,21 +26,115 @@ interface Props {
...
+          </div>`);
+        };
+        process();
+      } else {
+        // if no highlight, then no need to "process"
+        setHtml(docHtml ? DOMPurify.sanitize(docHtml, SANITIZE_CONFIG) : '');
+      }
+    }
+
+    setLoading(false);
```
maniax89: don't you want to wait until the above `process` command is complete to set loading to false? im not sure how long  `processDoc` takes but seems that you might want to wait to `setLoading` until after that is done

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -18,21 +26,115 @@ interface Props {
...
+  }, [document, highlight, setLoading]);
```
maniax89: i think the `useEffect` here will re-run a bunch for these non-primitive, non-react-generated input parameters. not sure if that is a concern for performance as I haven't tested it locally. one option is to use the `useDeepCompareEffect` which uses lodash's `isEqual` method on complex object types

regardless, this is a performance optimization only to revisit if we see things being slow, etc.

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -18,21 +26,115 @@ interface Props {
...
+
+    if (highlightNode) {
+      clearNodeChildren(highlightNode);
+    }
+
+    if (!highlight || !contentNode || !highlightNode) {
+      return;
+    }
+
+    const { begin, end } = get(highlight, 'table.location');
```
maniax89: should the fallback be `{}` here?

```suggestion
    const { begin, end } = get(highlight, 'table.location', {});
```

worried that trying a variable destructuring assignment will result in an error on `undefined`

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -18,21 +26,115 @@ interface Props {
...
+    });
+    highlightNode.appendChild(fragment);
 
-  setLoading(false);
+    // scroll highlight into view
+    const firstFieldRect = highlightNode.querySelector('.field--rect');
+    if (firstFieldRect) {
+      firstFieldRect.scrollIntoView({ block: 'center' });
+    }
+  }, [html, highlight]);
```
maniax89: same comment above about `useDeepCompareEffect`

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -18,21 +26,115 @@ interface Props {
...
+    'page'
+  ],
+  WHOLE_DOCUMENT: true
+};
+
+const base = `${settings.prefix}--html`;
+
+export const HtmlView: FC<Props> = ({
+  document,
+  highlight,
```
maniax89: what is `highlight` in this context? is it the `table_results` data?

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
jhpedemonte: Good catch.

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
jhpedemonte: So far, haven't seen perf issues.

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
jhpedemonte: Another good catch!

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
jhpedemonte: Same answer.

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
jhpedemonte: Yes.

PR #52, opened 2020-01-14
Merge january fix pack

maniax89:


packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -1,13 +1,21 @@
-import React, { FC } from 'react';
-import { QueryResult } from 'ibm-watson/discovery/v2';
+import React, { FC, useEffect, useState, useRef } from 'react';
+import { settings } from 'carbon-components';
+import { QueryResult, QueryTableResult } from 'ibm-watson/discovery/v2';
 import DOMPurify from 'dompurify';
 import get from 'lodash/get';
+import { processDoc } from 'utils/document/processDoc';
```
maniax89: i had to change this from `import processDoc from ''` to `import { processDoc } from ''` to get this to build - little worried why it worked before but not when merging back into `develop` -> @jhpedemonte any ideas here?

jhpedemonte: That doesn't make sense. It's exported as `default`:

```javascript
export default async function processDoc
```

so former should be what we need. We'll take a look.

maniax89: I think it might have been changed in https://github.com/watson-developer-cloud/discovery-components/pull/15

PR #53, opened 2020-01-14
chore: unversion discovery search app

maniax89:

stops versioning the `discovery-search-app` (which pollutes our github releases)


examples/discovery-search-app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "discovery-search-app",
-  "version": "1.0.2-beta.19",
+  "version": "1.0.0",
```
Steven-Gassert: is there any point to keeping around a version?

PR #55, opened 2020-01-14
feat: allow custom document text field name

ColbyJohnIBM:

Pulls text field name from context
https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/224

Check that documents without a `text` field are handled properly (not just a blank Document tab).


packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -44,7 +46,8 @@ export const SimpleDocument: FC<Props> = ({
     if (isJsonType && (!highlight || !isPassage(highlight))) {
       html = `<p>${cannotPreviewMessage}</p>`;
     } else {
-      let text = get(document, 'text', '');
+      const field = get(componentSettings, 'fields_shown.body.field', 'text');
```
jhpedemonte: Add a comment about what's going on here in these 2 lines.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/__tests__/SimpleDocument.test.tsx
```diff
@@ -45,4 +47,41 @@ describe('SimpleDocument', () => {
...
+            setHideToolbarControls={(): void => {}}
+            setLoading={(): void => {}}
+          />,
+          api,
+          context
+        )
+      ));
+    });
+
+    getByText(documentText);
```
maniax89: this isn't really testing anything - can you add an `expect` here to validate something?

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -34,6 +35,7 @@ export const SimpleDocument: FC<Props> = ({
 }) => {
   const contentRef = useRef<HTMLDivElement>(null);
   const highlightRef = useRef<HTMLDivElement>(null);
+  const { componentSettings } = React.useContext(SearchContext);
```
maniax89: could just import `useContext` above from react instead of doing `React.` -> i think this stuff only affects storybook occasionally but probably doesn't matter here

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/__tests__/SimpleDocument.test.tsx
```diff
@@ -45,4 +47,41 @@ describe('SimpleDocument', () => {
...
+    const documentText = 'This is the text of the document.';
+    const customDoc = {
+      body_field: documentText
+    };
+    const context: Partial<SearchContextIFC> = {
+      componentSettings: {
+        fields_shown: {
+          body: {
+            field: 'body_field',
+            use_passage: true
```
maniax89: i guess this code path doesn't care what `use_passage` is set to (in tooling, `body.field` is ignored in the search results if `use_passage` is `true` -> does the `DocumentPreview` still use the field from `document_passage[0].field` when highlighting?)

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/__tests__/SimpleDocument.test.tsx
jhpedemonte: The `get...` queries of `react-testing-library` do implicit assertions and will fail the test if they can't find the requested data. That lib also provides the same queries with `query...` instead -- these don't assert.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/__tests__/SimpleDocument.test.tsx
jhpedemonte: For passage highlighting, we'll pull the field from the `passage` data (or, now, from the `body.field`). Is that what you're asking?

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/__tests__/SimpleDocument.test.tsx
maniax89: So that throws an error if it fails, but does it show a success if it doesn't?

jhpedemonte: Well, the test succeeds.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/__tests__/SimpleDocument.test.tsx
maniax89: Right, in saying when use_passage is `true` you should always pull from the field specified in the passage data, even if body field is specified

jhpedemonte: Ah, right, I get you. Not only `body.use_passage`, but if we have a `passage` passed in, use that first; **then**, look in `body.field || text`.

Current code has issue that if `passage.field` is "text", we may instead be using `body.field`.

jhpedemonte: @ColbyJohnIBM I added a test which tests if passage field takes precedence over body field. That failed, so I moved the code around to handle this issue, too. Take a look: https://github.com/watson-developer-cloud/discovery-components/pull/55/commits/58aa82978a8dca4a7bf4708a4b3bb51787671f96

ColbyJohnIBM:
Ah, didn't realize about the passage piece. New code looks good to me. :+1:

PR #56, opened 2020-01-16
fix: simpleDoc passage highlighting

JuanNicolasGarcia:

This fixes the passage highlighting in `SimpleDocument` when the highlight location is more than 64KB worth of characters. 

Run storybook and go to DocumentPreview.

Select the passage highlighting story and select the multiline knob

See that `32.7 Buyer’s and Customer's Regulatory Authorities shall have the benefit of any rights of audit and access to information and documentation provided for in this Agreement to the extent that they relate to the exercise of the Regulatory Authorities' legal rights and/or responsibilities.` is highlighted.


JuanNicolasGarcia:
https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/226

JuanNicolasGarcia:
Unfortunately we are having difficultly making a test for this issue, since `react-testing-library` does not seem to allow us to break up the large textNode being made by the `SimpleDocument` component.  It would need to be broken up to properly test the offsets to be sure this is working test-wise.

I have tested this fix manually using storybook and using `yarn link` to link disco-components to disco-tooling and test against a cluster, locally.  After testing it, I have ran into no issues.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -44,6 +45,7 @@ export function findOffsetInDOM(
 
   do {
     endNode = allNodes[idx];
+    endNode.normalize();
```
jhpedemonte: The 2 calls shouldn't be here: we'd be calling `normalize()` unnecessarily, on nodes that we don't care about. We only need this in `getTextNodeAndOffset`

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -72,6 +74,7 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
   let textNode: Text,
     runningOffset = nodeOffset,
     len: number;
+
```
jhpedemonte: Call `node.normalize()` in this function. Afterwards, we shouldn't need the following `do/while` loop any more, as there will be a single TextNode. Actually, I don't think we need the `createNodeIterator` either.

jhpedemonte:
Closing in favor of #57

PR #57, opened 2020-01-16
fix: calculate DOM offset first, followed by TextNode

jhpedemonte:

In cases where an HTML tag has greater than 64K characters, the DOM in some browsers will break that up into multiple TextNodes. This fixes the offset calculation, so we can properly highlight text in those cases.

- Storybook
- Test


jhpedemonte:
@broulaye is working on a test. Will add that later today.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -99,13 +89,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
...
+  }
+
+  // In certain cases (i.e. text length > 65536 chars), the DOM will break
+  // up text into multiple TextNodes. Loop through TextNodes, finding which
+  // contains the `domOffset`
+  const iterator = document.createNodeIterator(node, NodeFilter.SHOW_TEXT);
+  let textNode: Text, len: number;
+  do {
+    textNode = iterator.nextNode() as Text;
+  } while (textNode && (len = textNode.data.length) && domOffset > len && (domOffset -= len));
```
ColbyJohnIBM: I still don't love the assignments being done in the `while` condition, but those have been around for a while...

ColbyJohnIBM:
Code looks good. Checking functionality locally.

packages/discovery-react-components/src/utils/document/documentUtils.ts
jhpedemonte: It's the best I could think of where the initialization happens both before and during, and the checks need to happen after that. Suggestions for improvement welcome.

ColbyJohnIBM: That's fair, any way I can think of to do it differently would lead to less clean code.

jhpedemonte:
Added test.

packages/discovery-react-components/src/utils/document/__tests__/documentUtils.spec.tsx
```diff
@@ -152,4 +153,27 @@ describe('findOffsetInDOM', () => {
     expect(beginOffset).toEqual(734);
     expect(endOffset).toEqual(107);
   });
+
+  it.only('finds offset when text spans multiple TextNodes', () => {
```
maniax89: doesn't this prevent other tests from running?

jhpedemonte: D'oh! Developer error. Good catch.

jhpedemonte:
Need to rebase on to `develop`.

JuanNicolasGarcia:
Rebased to `develop` now that `january_fix_pack` has been merged into develop.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -99,13 +89,26 @@ export function getTextNodeAndOffset(node: Node, offset: number): NodeOffset {
...
+
+  if (textNode === null) {
+    throw new Error(`Failed to find text node. Node: ${node.textContent}, offset: ${strOffset}`);
```
maniax89: is this message going to be displayed to the user or just logged internally? would be good to provide some i18n message text if its an end-user string

jhpedemonte: Not end-user visible, so don't think it requires i18n. Will show in browser console, though; may help with debugging an issue if we can ask a user to look for errors/warnings there.

PR #59, opened 2020-01-16
feat: implement fields api and use it in StructuredQuery

Steven-Gassert:
This implements the `useFieldsApi` in order to fetch project fields and display them in our StructuredQuery component.

https://github.ibm.com/Watson-Discovery/disco-widgets/issues/184

You can verify this change by
1) Making sure the new unit tests added to `useDataApi.test.tsx`, `StructuredQuery.test.tsx`, and `DiscoverySearch.test.tsx` all pass and capture the intended functionality.
2) Making sure that the fields dropdown in storybook works as expected.
3) adding `StructuredQuery` to our example app (haven't added it yet since it's functionality is not complete) and checking that fields populate in the **Fields** dropdown menu

I've added this component to our storybook
No

### DIFF
--- a/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
+++ b/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, FC, useEffect, useState, useCallback, useMemo } from 'react';
+import React, { createContext, FC, useEffect, useState, useMemo, useCallback } from 'react';
 import DiscoveryV2 from 'ibm-watson/discovery/v2';
 import {
   useDeepCompareEffect,
@@ -11,7 +11,9 @@ import {
   AutocompleteStore,
   useSearchResultsApi,
   useFetchDocumentsApi,
-  useAutocompleteApi
+  useAutocompleteApi,
+  useFieldsApi,
+  FieldsStore
 } from 'utils/useDataApi';
 import { SearchClient } from './types';
 import { withErrorBoundary } from 'react-error-boundary';
@@ -99,6 +101,7 @@ export interface SearchContextIFC {
   autocompletionStore: AutocompleteStore;
   componentSettings: DiscoveryV2.ComponentSettingsResponse | null;
   isResultsPaginationComponentHidden: boolean | undefined;
+  fieldsStore: FieldsStore;
 }
 
 export interface SearchApiIFC {
@@ -116,6 +119,7 @@ export interface SearchApiIFC {
   setIsResultsPaginationComponentHidden: (
     isResultsPaginationComponentHidden: boolean | React.SetStateAction<boolean | undefined>
   ) => void;
+  fetchFields: () => void;
 }
 
 export const searchApiDefaults = {
@@ -127,7 +131,8 @@ export const searchApiDefaults = {
   setSelectedResult: (): void => {},
   setAutocompletionOptions: (): void => {},
   setSearchParameters: (): void => {},
-  setIsResultsPaginationComponentHidden: (): void => {}
+  setIsResultsPaginationComponentHidden: (): void => {},
+  fetchFields: (): Promise<void> => Promise.resolve()
 };
 
 export const searchResponseStoreDefaults: SearchResponseStore = {
@@ -175,6 +180,15 @@ const aggregationQueryDefaults: Partial<DiscoveryV2.QueryParams> = {
   }
 };
 
+const fieldsStoreDefaults: FieldsStore = {
+  parameters: {
+    projectId: ''
+  },
+  data: null,
+  isLoading: false,
+  isError: false
+};
+
 export const searchContextDefaults = {
   aggregationResults: null,
   searchResponseStore: searchResponseStoreDefaults,
@@ -183,7 +197,8 @@ export const searchContextDefaults = {
   autocompletionStore: autocompletionStoreDefaults,
   collectionsResults: null,
   componentSettings: null,
-  isResultsPaginationComponentHidden: false
+  isResultsPaginationComponentHidden: false,
+  fieldsStore: fieldsStoreDefaults
 };
 
 export const SearchApi = createContext<SearchApiIFC>(searchApiDefaults);
@@ -396,6 +411,12 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
     [fetchDocuments, setSearchResponse]
   );
 
+  const [fieldsStore, { fetchFields }] = useFieldsApi({ projectId }, searchClient);
+
+  const handleFetchFields = useCallback(() => {
+    fetchFields();
+  }, [fetchFields]);
+
   const api = useMemo((): SearchApiIFC => {
     return {
       performSearch: handleSearch,
@@ -405,14 +426,16 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
       setSelectedResult: handleSetSelectedResult,
       setAutocompletionOptions,
       setSearchParameters,
-      setIsResultsPaginationComponentHidden
+      setIsResultsPaginationComponentHidden,
+      fetchFields: handleFetchFields
     };
   }, [
+    handleSearch,
     handleFetchAggregations,
     handleFetchAutocompletions,
-    handleSearch,
     handleFetchDocuments,
-    setSearchParameters
+    setSearchParameters,
+    handleFetchFields
   ]);
 
   const state = useDeepCompareMemo(() => {
@@ -424,7 +447,8 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
       selectedResult,
       collectionsResults,
       componentSettings,
-      isResultsPaginationComponentHidden
+      isResultsPaginationComponentHidden,
+      fieldsStore
     };
   }, [
     aggregationResults,
@@ -434,7 +458,8 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
     selectedResult,
     collectionsResults,
     componentSettings,
-    isResultsPaginationComponentHidden
+    isResultsPaginationComponentHidden,
+    fieldsStore
   ]);
 
   return (
--- a/packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
+++ b/packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState } from 'react';
+import React, { FC, useState, useContext, useEffect } from 'react';
 import { RuleGroup } from './components/RuleGroup/RuleGroup';
 import { AddRuleRowButton } from './components/AddRuleRowButton/AddRuleRowButton';
 import { AddRuleGroupButton } from './components/AddRuleGroupButton/AddRuleGroupButton';
@@ -9,6 +9,7 @@ import { StructuredQuerySelection } from './utils/structuredQueryInterfaces';
 import { withErrorBoundary } from 'react-error-boundary';
 import { FallbackComponent } from 'utils/FallbackComponent';
 import onErrorCallback from 'utils/onErrorCallback';
+import { SearchApi } from 'components/DiscoverySearch/DiscoverySearch';
 
 export interface StructuredQueryProps {
   /**
@@ -32,6 +33,11 @@ const StructuredQuery: FC<StructuredQueryProps> = ({ messages = defaultMessages 
     structuredQuerySelection.groups[0].rows.length < MAX_NUM_SIBLING_RULE_ROWS;
   const showAddRuleGroupButton =
     Object.keys(structuredQuerySelection.groups).length - 1 < MAX_NUM_NESTED_RULE_GROUPS;
+
+  const { fetchFields } = useContext(SearchApi);
+  useEffect(() => {
+    fetchFields();
+  }, [fetchFields]);
 
   return (
     <div className={structuredQueryClass}>
--- a/packages/discovery-react-components/src/components/StructuredQuery/__stories__/StructuredQuery.stories.tsx
+++ b/packages/discovery-react-components/src/components/StructuredQuery/__stories__/StructuredQuery.stories.tsx
@@ -7,13 +7,25 @@ import StructuredQuery from '../StructuredQuery';
 import { defaultMessages } from '../messages';
 import marked from 'marked';
 import defaultReadme from './default.md';
+import { createDummyResponsePromise } from 'utils/testingUtils';
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
+import { action } from '@storybook/addon-actions';
 
 const props = () => ({
   messages: object("Default messages for the component's text strings", defaultMessages)
 });
 
+class DummySearchClientReturnsFields extends DummySearchClient {
+  public async listFields(fetchFieldsParams: DiscoveryV2.ListFieldsParams): Promise<any> {
+    action('listFields')(fetchFieldsParams);
+    return createDummyResponsePromise({
+      fields: [{ field: 'field name 1' }, { field: 'field name 2' }]
+    });
+  }
+}
+
 const discoverySearchProps = (): DiscoverySearchProps => ({
-  searchClient: new DummySearchClient(),
+  searchClient: new DummySearchClientReturnsFields(),
   projectId: text('Project ID', 'project-id')
 });
 
--- a/packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
+++ b/packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
@@ -1,16 +1,79 @@
 import React from 'react';
-import { render, RenderResult } from '@testing-library/react';
-import StructuredQuery from '../StructuredQuery';
+import { render, RenderResult, fireEvent } from '@testing-library/react';
+import StructuredQuery, { StructuredQueryProps } from '../StructuredQuery';
+import { wrapWithContext } from 'utils/testingUtils';
+import {
+  SearchContextIFC,
+  searchResponseStoreDefaults,
+  fetchDocumentsResponseStoreDefaults,
+  SearchApiIFC
+} from 'components/DiscoverySearch/DiscoverySearch';
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
+import '@testing-library/jest-dom/extend-expect';
+
+interface SetupParams {
+  projectFields?: DiscoveryV2.Field[];
+  fieldsStoreLoadingState?: boolean;
+  fieldsStoreErrorState?: boolean;
+  apiOverrides?: Partial<SearchApiIFC>;
+}
+interface Setup {
+  structuredQuery: RenderResult;
+}
+
+function setup(
+  {
+    projectFields,
+    fieldsStoreLoadingState = false,
+    fieldsStoreErrorState = false,
+    apiOverrides = {}
+  }: SetupParams,
+  componentProps: Partial<StructuredQueryProps> = {}
+): Setup {
+  let context: Partial<SearchContextIFC> = {
+    searchResponseStore: {
+      ...searchResponseStoreDefaults
+    },
+    fetchDocumentsResponseStore: {
+      ...fetchDocumentsResponseStoreDefaults
+    },
+    fieldsStore: {
+      data: { fields: projectFields },
+      isLoading: fieldsStoreLoadingState,
+      isError: fieldsStoreErrorState,
+      parameters: {
+        projectId: ''
+      }
+    }
+  };
+  const structuredQuery = render(
+    wrapWithContext(<StructuredQuery {...componentProps} />, apiOverrides, context)
+  );
+  return {
+    structuredQuery
+  };
+}
 
 describe('<StructuredQuery />', () => {
+  let structuredQuery: RenderResult;
+  let projectFields: DiscoveryV2.Field[] | undefined = undefined;
+  let fieldsStoreLoadingState: boolean;
+  let fieldsStoreErrorState: boolean;
+
+  afterEach(() => {
+    projectFields = undefined;
+    fieldsStoreLoadingState = false;
+    fieldsStoreErrorState = false;
+  });
+
   describe('i18n messages', () => {
     describe('when no messages are provided', () => {
-      let structuredQuery: RenderResult;
-      beforeEach(() => {
-        structuredQuery = render(<StructuredQuery />);
-      });
-
       describe('has the correct default text', () => {
+        let structuredQuery: RenderResult;
+        beforeEach(() => {
+          structuredQuery = render(<StructuredQuery />);
+        });
+
         test('rule group dropdown has the correct default messages', () => {
           const ruleGroupDropdownTextOne = structuredQuery.getByText('Satisfy', {
             exact: false
@@ -29,12 +92,6 @@ describe('<StructuredQuery />', () => {
           expect(addRuleRowText).toBeDefined();
           expect(addGroupRulesText).toBeDefined();
         });
-        test('field dropdown has the correct default messages', () => {
-          const fieldDropdownTitleText = structuredQuery.getByText('Field');
-          const fieldDropdownPlaceholderText = structuredQuery.getByPlaceholderText('Select field');
-          expect(fieldDropdownTitleText).toBeDefined();
-          expect(fieldDropdownPlaceholderText).toBeDefined();
-        });
         test('operator dropdown has the correct default messages', () => {
           const operatorDropdownTitleText = structuredQuery.getByText('Operator');
           const operatorDropdownPlaceholderText = structuredQuery.getByPlaceholderText(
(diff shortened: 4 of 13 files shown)


packages/discovery-react-components/src/utils/__tests__/useDataApi.test.tsx
```diff
@@ -40,20 +47,38 @@ class SingleQueryResultSearchClient extends BaseSearchClient {
   }
 }
 
+// class SingleFieldResponseSearchClient extends BaseSearchClient {
```
Steven-Gassert: will remove

packages/discovery-react-components/src/utils/__tests__/useDataApi.test.tsx
```diff
@@ -8,9 +8,12 @@ import {
   useFetchDocumentsApi,
   FetchDocumentsResponseStore,
   useAutocompleteApi,
-  AutocompleteStore
+  AutocompleteStore,
+  useListFieldsApi,
+  ListFieldsStore
 } from '../useDataApi';
 import { SearchClient } from 'components/DiscoverySearch/types';
+import { isEqual } from 'lodash';
```
Steven-Gassert: will remove

packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -234,6 +278,7 @@ export interface FetchDocumentsResponseStore extends ReducerState {
 /**
  * fetch documents actions used to interact with the search API and search state
  */
+//here
```
Steven-Gassert: will remove

packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -162,6 +162,50 @@ interface ReducerState {
...
+}
+
+/**
+ * concrete usage of the useDataApi helper method for fetching project fields
+ * @param searchParameters - initial search parameters to set
+ * @param searchClient - search client used to perform requests
+ * @return a 2-element array containing the list fields store data and listFields-specific store actions
+ */
+export const useListFieldsApi = (
+  searchParameters: DiscoveryV2.QueryParams,
```
maniax89: shouldn't this be `DiscoveryV2.ListFieldsParams` ?

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -388,4 +448,58 @@ describe('<StructuredQuery />', () => {
       });
     });
   });
+
+  describe('On mount', () => {
+    it.only('Calls list fields', () => {
```
kabeaty: I guess this `.only` should go away :)

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -388,4 +448,58 @@ describe('<StructuredQuery />', () => {
...
+    beforeEach(() => {
+      projectFields = undefined;
+    });
+
+    it('should display the loading text', () => {
+      ({ structuredQuery } = setup({ projectFields }));
+      expect(
+        structuredQuery.getByPlaceholderText(defaultMessages.fieldDropdownLoadingText)
+      ).toBeInTheDocument();
+    });
```
kabeaty: Should we also check that it's disabled at this point?

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
```diff
@@ -87,5 +91,6 @@ export const defaultMessages: Messages = {
   valueInputPlaceholderText: 'Enter value',
   addRuleRowText: 'Add rule',
   removeRuleRowButtonIconDescription: 'Remove rule',
-  addRuleGroupText: 'Add group of rules'
+  addRuleGroupText: 'Add group of rules',
+  fieldDropdownLoadingText: 'Fetching project fields'
```
kabeaty: `Fetching` seems more technical and less friendly than `Loading` to me for the default text here.

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -45,16 +53,26 @@ export const RuleRow: FC<RuleRowProps> = ({
   const showRemoveRuleRowButton =
     structuredQuerySelection.groups[groupId].rows.length > 1 || !isTopLevelGroup;
 
+  // TODO Check out missing prop warning on <ComboBox />
```
kabeaty: Is this comment about `items` being required while the `Combobox` is loading/disabled?

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
```diff
@@ -33,6 +35,12 @@ const StructuredQuery: FC<StructuredQueryProps> = ({ messages = defaultMessages
   const showAddRuleGroupButton =
     Object.keys(structuredQuerySelection.groups).length - 1 < MAX_NUM_NESTED_RULE_GROUPS;
 
+  const { listFields } = useContext(SearchApi);
+  useDeepCompareEffect(() => {
+    console.log('calling listFields from useEffect hook');
```
kabeaty: I guess you're planning to remove this too :)

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -57,6 +59,14 @@ export interface DiscoverySearchProps {
...
    */
   overrideComponentSettings?: DiscoveryV2.ComponentSettingsResponse;
+  /**
+   * overrideListFieldsParams is used to override internal listFieldsParams value. See the Discovery Node SDK ListFieldsParams type for more info
+   */
+  overrideListFieldsParams?: Partial<DiscoveryV2.ListFieldsParams>;
+  /**
+   * overrideProjectFields is used to override internal listFields state
+   */
+  overrideListFieldsResponse?: DiscoveryV2.ListFieldsResponse;
```
kabeaty: Maybe `overrideListFieldsResults` or `overrideFieldsResults` here for consistency with `overrideCollectionsResults` and 'overrideAutocompletionsResults` above?

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -45,16 +53,26 @@ export const RuleRow: FC<RuleRowProps> = ({
...
     <div className={structuredQueryRulesClass} data-testid={`rule-row-${groupId}`}>
-      <ComboBox
-        id={`structured-query-rules-field-${groupId}`}
-        // TODO: Items is empty for now as it's a required field and retrieving fields for the dropdown
-        // and adding them as items will be addressed in a future issue
-        items={[]}
-        placeholder={messages.fieldDropdownPlaceholderText}
-        titleText={messages.fieldDropdownTitleText}
-      />
+      {isLoading || !projectFields ? (
```
maniax89: could we inline the loading message in the `placeholder` attribute then have `projectFields` default to an array instead of this ternary?

packages/discovery-react-components/src/components/StructuredQuery/messages.ts
maniax89: nitpick - can we put this next to the other `field*` messages above?

packages/discovery-react-components/src/components/StructuredQuery/utils/getProjectFields.ts
```diff
@@ -0,0 +1,14 @@
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
+
+export const getProjectFields = (
+  response: DiscoveryV2.ListFieldsResponse | null
+): string[] | null => {
+  if (response && response.fields) {
+    let fieldObjects = response.fields;
+    fieldObjects = fieldObjects.filter(fieldObject => fieldObject.field);
```
maniax89: nit - couldn't this be `reduce` instead of looping twice?
```suggestion
    const fieldNames = response.fields.reduce((accumulator, fieldObject) => {
      if (fieldObject.field) {
        accumulator.push(fieldObject.field);
      }
      return accumulator;
    }, []);
```

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -57,6 +59,14 @@ export interface DiscoverySearchProps {
    * overrideComponentSettings is used to override internal collections result state
    */
   overrideComponentSettings?: DiscoveryV2.ComponentSettingsResponse;
+  /**
+   * overrideListFieldsParams is used to override internal listFieldsParams value. See the Discovery Node SDK ListFieldsParams type for more info
+   */
+  overrideListFieldsParams?: Partial<DiscoveryV2.ListFieldsParams>;
```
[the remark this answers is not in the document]
Steven-Gassert: it's a pretty small use case, but I was thinking that maybe someone would want to override the collection id's that this call would operate on? I'm totally cool with getting rid of this if you think that use case is too niche to worry about

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -116,6 +127,7 @@ export interface SearchApiIFC {
   setIsResultsPaginationComponentHidden: (
     isResultsPaginationComponentHidden: boolean | React.SetStateAction<boolean | undefined>
   ) => void;
+  listFields: () => void;
```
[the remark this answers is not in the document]
Steven-Gassert: I agree this would be a better name

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
Steven-Gassert: going to dive into this a bit later, but this prop warning was on the `onChange` prop

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
Steven-Gassert: added

packages/discovery-react-components/src/components/StructuredQuery/__stories__/StructuredQuery.stories.tsx
```diff
@@ -14,7 +14,10 @@ const props = () => ({
 
 const discoverySearchProps = (): DiscoverySearchProps => ({
   searchClient: new DummySearchClient(),
-  projectId: text('Project ID', 'project-id')
+  projectId: text('Project ID', 'project-id'),
+  overrideListFieldsResponse: {
```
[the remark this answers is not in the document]
Steven-Gassert: This makes more sense to me, I've made these changes

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -52,6 +106,36 @@ describe('<StructuredQuery />', () => {
...
 
+    describe('when there are fields present', () => {
+      beforeEach(() => {
+        projectFields = [{ field: 'field name 1' }, { field: 'field name 2' }];
+        ({ structuredQuery } = setup({ projectFields }));
+      });
+
+      test('field dropdown has the correct default messages', () => {
+        const fieldDropdownTitleText = structuredQuery.getByText('Field');
+        const fieldDropdownPlaceholderText = structuredQuery!.getByPlaceholderText('Select field');
```
maniax89: why does this need `!.`?

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -90,23 +174,18 @@ describe('<StructuredQuery />', () => {
             <StructuredQuery
               messages={{
                 addRuleGroupText: 'A new rules group',
-                fieldDropdownPlaceholderText: 'Field choice selection',
```
maniax89: why is this being removed? shouldn't the placeholder still exist until a selection is made?

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -388,4 +467,75 @@ describe('<StructuredQuery />', () => {
...
+      });
+    });
+  });
+
+  describe('when there are no fields available', () => {
+    beforeEach(() => {
+      projectFields = undefined;
+    });
+
+    it('should display the loading text', () => {
```
maniax89: i think it should only display the loading text if `isLoading` is `true` -> otherwise there may be the false understanding that eventually fields will be loaded even if no request is running

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -1,16 +1,76 @@
...
+  );
+  return {
+    structuredQuery
+  };
+}
 
 describe('<StructuredQuery />', () => {
+  let structuredQuery: RenderResult;
+  let projectFields: DiscoveryV2.Field[] | undefined = undefined;
+  let fieldsStoreLoadingState: boolean | undefined;
```
maniax89: shouldn't this just be a `boolean`? im not sure why it would ever be `undefined`

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
```diff
@@ -29,12 +89,6 @@ describe('<StructuredQuery />', () => {
           expect(addRuleRowText).toBeDefined();
           expect(addGroupRulesText).toBeDefined();
         });
-        test('field dropdown has the correct default messages', () => {
```
maniax89: why is this being removed?

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -45,15 +54,20 @@ export const RuleRow: FC<RuleRowProps> = ({
...
-        items={[]}
-        placeholder={messages.fieldDropdownPlaceholderText}
+        items={projectFields}
+        placeholder={
+          haveProjectFields
+            ? messages.fieldDropdownPlaceholderText
+            : messages.fieldDropdownLoadingText
+        }
         titleText={messages.fieldDropdownTitleText}
+        data-test-id={`structured-query-rules-field-${groupId}`}
```
maniax89: i don't see this being used in tests, but if it were, i would imagine that it should be `data-testid` instead

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
maniax89: did this get resolved?

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
```diff
@@ -45,15 +54,20 @@ export const RuleRow: FC<RuleRowProps> = ({
...
     <div className={structuredQueryRulesClass} data-testid={`rule-row-${groupId}`}>
       <ComboBox
         id={`structured-query-rules-field-${groupId}`}
-        // TODO: Items is empty for now as it's a required field and retrieving fields for the dropdown
-        // and adding them as items will be addressed in a future issue
-        items={[]}
-        placeholder={messages.fieldDropdownPlaceholderText}
+        items={projectFields}
+        placeholder={
+          haveProjectFields
```
maniax89: i think the check should be `isLoading` not `haveProjectFields` as we don't know whether or not fields will ever be fetched (kind of a false indicator)

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
```diff
@@ -0,0 +1,16 @@
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
+
+export const getFieldNames = (response: DiscoveryV2.ListFieldsResponse | null): string[] => {
+  if (response && response.fields) {
+    let fieldObjects = response.fields;
```
maniax89: nit - this can be a const

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
```diff
@@ -0,0 +1,16 @@
...
+    const fieldNames = fieldObjects.reduce((namesArray: string[], fieldObject) => {
+      if (fieldObject.field) {
+        namesArray.push(fieldObject.field);
```
maniax89: since you aren't checking the `collection_id`, you might want to make sure you aren't adding duplicate fields.

i would test this out by creating 2 collections and putting documents in both and then validating that the field names you get back from this function are unique (i don't think they will be)

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
```diff
@@ -0,0 +1,16 @@
...
+      }
+      return namesArray;
+    }, []);
+    return fieldNames;
+  } else {
```
maniax89: nit - don't need `else` here

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
maniax89: another thing - each field is going to have a `type` -> might want to revisit this later but i don't think all field types would be beneficial to include in this dropdown (namely, we should probably filter out `nested` types)

packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -162,6 +160,53 @@ interface ReducerState {
...
+  const { state: filedsState, setData: setFieldsResponse, setFetchToken } = useDataApi(
+    fetchFieldsParams,
+    null,
+    searchClient.listFields,
+    searchClient
+  );
+  return [
+    filedsState,
+    {
+      fetchFields: (callback?: (results: DiscoveryV2.ListFieldsResponse) => void) => {
```
maniax89: i don't think you are using the `callback` here, so its just added complexity that can be removed. i think this can just be
```suggestion
      fetchFields: () => {
        setFetchToken({ trigger: true })
      }
```

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
Steven-Gassert: Not yet, circling back on this now

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
Steven-Gassert: previously I was reseting `structuredQuery` to null after every test to avoid accidentally adding elements to an already rendered tree, which I since realized was unnecessary. While the value of `structuredQuery` could have been null, I needed to cast this, but I should be fine to remove this now thanks for the catch

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
Steven-Gassert: I previously was following the pattern of resetting my component params to `undefined` after each test to bring us back to the same `undefined` state after each test. But you're right that it probably makes sense to set this to a default value of `false` here

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
Steven-Gassert: This test was placed under the `when there are no fields present` block

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
Steven-Gassert: I originally removed this because in this case we have not setup our DummySearchClient to return fields, thus the loading text should be displayed instead of the placeholder text. I've since updated this test to check for the existence of the loading text instead of just removing the placeholder text

packages/discovery-react-components/src/components/StructuredQuery/__tests__/StructuredQuery.test.tsx
Steven-Gassert: I could see how this could be a better experience. Maybe I can also add an `fieldDropdownErrorText` to display if the fields store has a true value for `isError`

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
Steven-Gassert: After some more thought, this does sound like a better strategy and I've made the corresponding changes

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
Steven-Gassert: I've removed this id that's no longer needed

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
Steven-Gassert: Ah interesting, nice catch. I'll test this out and can also check out what they did for this dropdown in the search results improvement tools section of the Tooling app

Steven-Gassert: it was easy enough to filter out fields with type `nested`. I also noticed that in the Tooling repo they were making sure the array of field names were unique, which confirms what you had thought might be the behavior when there are multiple collections with identical field names, so I also utilized `uniq` in our helper before sending this array of field names back to the `RuleRow` component

packages/discovery-react-components/src/components/StructuredQuery/components/RuleRow/RuleRow.tsx
Steven-Gassert: I see that the `onChange` is a param meant for you to feed a callback and make this component fully controlled. I imagine we'll make use of `onChange` and the `selectedItem` params when were ready to make this component controlled and start assembling our queries. I've inserted a empty function function for now as well as some comments to come back and change this in the issue that tackles assembling the query

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
```diff
@@ -0,0 +1,18 @@
...
+    const fieldObjects = response.fields;
+    const fieldNames = fieldObjects.reduce((namesArray: string[], fieldObject) => {
+      if (fieldObject.type === 'nested') {
+        return namesArray;
+      } else if (fieldObject.field) {
+        namesArray.push(fieldObject.field);
+      }
+      return namesArray;
+    }, []);
+    return uniq(fieldNames);
```
maniax89: nit - you can just do `else if (fieldObject.field && !namesArray.includes(fieldObject.field)) {` instead of doing a uniq check

packages/discovery-react-components/src/components/StructuredQuery/utils/__tests__/getFieldNames.test.ts
```diff
@@ -0,0 +1,49 @@
+import { getFieldNames } from '../getFieldNames';
+
+function matchingArrays(array1: string[], array2: string[]): boolean {
```
maniax89: nit - since these are arrays of primitives, you could just compare the strings -> array1.toString() === array2.toString()

packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -162,6 +160,53 @@ interface ReducerState {
...
+ * @param searchParameters - initial search parameters to set
+ * @param searchClient - search client used to perform requests
+ * @return a 2-element array containing the fields store data and fields-specific store actions
+ */
+
+export const useFieldsApi = (
+  fetchFieldsParams: DiscoveryV2.ListFieldsParams,
+  searchClient: SearchClient
+): [FieldsStore, FieldsStoreActions] => {
+  const { state: filedsState, setData: setFieldsResponse, setFetchToken } = useDataApi(
```
ChrisGillin: Should this be `state: fieldsState`?

packages/discovery-react-components/src/components/StructuredQuery/utils/getFieldNames.ts
Steven-Gassert: Ah good call, thanks

packages/discovery-react-components/src/components/StructuredQuery/utils/__tests__/getFieldNames.test.ts
Steven-Gassert: another good call, thanks!

packages/discovery-react-components/src/utils/useDataApi.ts
Steven-Gassert: Yup, thanks for the catch

Steven-Gassert:
going to close and open this PR to trigger a rebuild on Travis

PR #62, opened 2020-01-20
Fixpack to develop

maniax89:


examples/discovery-search-app/package.json
```diff
@@ -21,8 +21,8 @@
   },
   "dependencies": {
     "@carbon/icons": "^10.5.0",
-    "@ibm-watson/discovery-react-components": "^1.1.0-beta.15",
-    "@ibm-watson/discovery-styles": "^1.1.0-beta.6",
+    "@ibm-watson/discovery-react-components": "^1.0.7-rc.1",
+    "@ibm-watson/discovery-styles": "^1.0.7-rc.1",
```
jhpedemonte: Update to `1.1.0-beta...` right?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -43,10 +43,10 @@ import {
   Field,
   Attributes,
   Relations
-} from 'components/CIDocument/types';
+} from '../../types';
 import { Address, Mention } from '../MetadataPane/types';
 import { Items } from '../DetailsPane/types';
-import { defaultTheme, Theme } from 'utils/theme';
+import { defaultTheme, Theme } from '../../../../utils/theme';
```
jhpedemonte: Are these regressions? Didn't we add code to avoid having to do these relative paths?

anibapat: I think `january_fix_pack` branch did not have the fix which is there in develop which prevents the relative path, that's why the relative paths are there in fixpack branch, and now it is coming on top of `develop` again.

PR #68, opened 2020-01-22
feat: add ci document to example app

broulaye:

Add CI Document to example app and add cypress test

Run Cypress test

Yes
No

https://zenhub.ibm.com/app/workspaces/discovery-unified-5728bf39d2c20f776e05d0ef/issues/watson-discovery/docviz-squad-issue-tracker/232

### DIFF
--- a/packages/discovery-react-components/src/components/CIDocument/__stories__/default.md
+++ b/packages/discovery-react-components/src/components/CIDocument/__stories__/default.md
@@ -1,3 +1,3 @@
 #### Overview
 
-The `CIDocument` component provides Element Classification to the user. It takes a PDF and enriches it with features a customer might deem important (such as Obligation, Deliverables, Assignments, etc...). This would help customers get a better understanding of their document, or help them find what they need in a large document in a more efficient way.
+The `CIDocument` component provides the user with a document enriched by one of the content intelligence enrichments. It displays a document with features a customer might deem important (such as Obligation, Deliverables, Assignments, etc...). This would help customers get a better understanding of their document, or help them find what they need in a large document in a more efficient way.


examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -101,5 +101,44 @@ describe('Passage Results', () => {
       cy.get('@combinedResult').contains('View passage in document');
       cy.get('@combinedResult').contains('View table in document');
     });
+
+    describe('and clicking on "CI Document" for a result', () => {
+      beforeEach(() => {
+        cy.get('button[data-testid="search-result-element-preview-button"]')
```
maniax89: is there button text you can look for here?

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -101,5 +101,44 @@ describe('Passage Results', () => {
...
+          .last()
+          .click();
+      });
+
+      it('verify that CI Document loaded correctly', () => {
```
maniax89: nice work! are there any more specifics around "loaded correctly" that could be split into separate `it` statements? seems like a good split might be "it can filter", "it has navigation tabs" or something similar

examples/discovery-search-app/src/App.js
```diff
@@ -5,6 +5,8 @@ import './app.scss';
 import { settings } from 'carbon-components';
 import { Button } from 'carbon-components-react';
 import Close from '@carbon/icons-react/lib/close/16';
+import { Tabs, Tab } from 'carbon-components-react';
```
maniax89: can you move this to be combined with `import { Button } from 'carbon-components-react';` above?

examples/discovery-search-app/src/App.js
```diff
@@ -85,9 +89,28 @@ function SearchPage() {
...
+    }
+    return [];
+  }
+
+  const tabs = [
+    {
+      name: 'Document',
+      Component: DocumentPreview
+    },
+    // Check if Content Intelligence (chewy) feature is enabled
```
maniax89: probably remove "chewy" here :-)

examples/discovery-search-app/src/App.js
```diff
@@ -85,9 +89,28 @@ function SearchPage() {
   );
 }
 
-function PreviewPage() {
+function PreviewPage(props) {
   const { setSelectedResult } = useContext(SearchApi);
 
+  function doRenderTab(testResult, content) {
```
maniax89: when looping through `tabs`, wouldn't `[]` still trigger an item in the array?

i feel like a similar approach would just be
```
const tabs = [
    {
      name: 'Document',
      Component: DocumentPreview
    }
];

if (canRenderCIDocument(props.document)) {
  tabs.push({ name: 'CI', Component: CIDocument });
}
```

examples/discovery-search-app/src/app.scss
```diff
@@ -129,4 +128,18 @@ $nav-toolbar-height: 42px;
     margin-top: $nav-toolbar-height;
     padding: $spacing-xl;
   }
+
+  .#{$prefix}--search-app__tabs {
+    flex: 0 0 auto;
+  }
+
+  .#{$prefix}--search-app__tabs-hidden {
```
maniax89: nitpick - `--hidden` for BEM syntax

examples/discovery-search-app/cypress/integration/passages/spec.ts
H-Fin: Very true! Just added some clearer sections.

H-Fin:
@maniax89 This is ready for a re-review

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -99,5 +99,50 @@ describe('Passage Results', () => {
...
+        cy.get('button[data-testid="search-result-element-preview-button"]')
+          .contains('View table in document')
+          .click();
+        cy.get('ul')
+          .contains('CI')
+          .should('exist');
+        cy.get('ul')
+          .contains('CI')
+          .click();
+      });
```
jhpedemonte: Doing gets on specific tags is fragile: if the underlying implementation changes from `ul` to `div`, for example, while keeping user-facing stuff basically the same, then this test will fail.

If there is no unique text to latch on to, then this looks like a good place to add a `data-testid`.

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -99,5 +99,50 @@ describe('Passage Results', () => {
...
+
+      it('verify that CI Document sections are present', () => {
+        cy.get('.bx--document-preview').should('exist');
+        cy.get('.bx--ci-doc__sidebar').should('exist');
+        cy.get('.bx--ci-doc__doc').should('exist');
+        cy.get('.bx--ci-doc__details').should('exist');
```
jhpedemonte: Again here, this seems fragile (although it looks like others are doing it as well). I really like the `react-testing-library` ethos, wherein the test should try to only use user-visible constructs (text, buttons, links) or `data-testid` as a last resort.

I believe we've done similar checks in our unit tests where we just looked for the appropriate (unique) text. We should do the same here and below.

examples/discovery-search-app/src/App.js
```diff
@@ -85,9 +88,20 @@ function SearchPage() {
   );
 }
 
-function PreviewPage() {
+function PreviewPage(props) {
```
jhpedemonte: -> `({ document })`, therefore you won't need to do `props.document` several times below.

broulaye:
Ran into some issues had to create a different PR here: https://github.com/watson-developer-cloud/discovery-components/pull/71

PR #69, opened 2020-01-22
fix: transformEnrichments takes an array and returns an array

anibapat:

This PR fixes the problem where TransformEnrichment returns an object instead of an array.

https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/225

~~Also, upgrades `typescript` and `prettier` version to the latest so that it can support es2020 features like optional chaining.~~

yarn test, 
Also improved test coverage for transformEnrichment.  

No

No

### DIFF
--- a/packages/discovery-react-components/src/components/CIDocument/types.ts
+++ b/packages/discovery-react-components/src/components/CIDocument/types.ts
@@ -46,7 +46,11 @@ export interface ItemMap {
   bySection: any;
 }
 
-export type EnrichedHtml = EnrichedHtmlContract | EnrichedHtmlInvoice | EnrichedHtmlPurchaseOrder;
+export type EnrichedHtml =
+  | EnrichedHtmlContract
+  | EnrichedHtmlInvoice
+  | EnrichedHtmlPurchaseOrder
+  | any;
 
 interface EnrichedHtmlContract {
   contract: Contract;
--- a/packages/discovery-react-components/src/components/CIDocument/types.ts
+++ b/packages/discovery-react-components/src/components/CIDocument/types.ts
@@ -46,11 +46,7 @@ export interface ItemMap {
   bySection: any;
 }
 
-export type EnrichedHtml =
-  | EnrichedHtmlContract
-  | EnrichedHtmlInvoice
-  | EnrichedHtmlPurchaseOrder
-  | any;
+export type EnrichedHtml = EnrichedHtmlContract | EnrichedHtmlInvoice | EnrichedHtmlPurchaseOrder;
 
 interface EnrichedHtmlContract {
   contract: Contract;
--- a/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
+++ b/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
@@ -5,7 +5,7 @@ import invoiceData from 'components/CIDocument/components/CIDocument/__fixtures_
 import purchaseOrderData from 'components/CIDocument/components/CIDocument/__fixtures__/po-index_op.json';
 import { EnrichedHtml } from 'components/CIDocument/types';
 
-describe('transformEnrichment', () => {
+describe('transformEnrichment - contracts', () => {
   let transformedDoc: EnrichedHtml[];
   const enriched_html = contractData.enriched_html;
 
@@ -31,7 +31,7 @@ describe('transformEnrichment', () => {
   });
 });
 
-describe('transformEnrichment', () => {
+describe('transformEnrichment - Invoice', () => {
   let transformedDoc: EnrichedHtml[];
   const enriched_html = invoiceData.enriched_html;
 
@@ -56,7 +56,7 @@ describe('transformEnrichment', () => {
   });
 });
 
-describe('transformEnrichment', () => {
+describe('transformEnrichment - Purchase orders', () => {
   let transformedDoc: EnrichedHtml[];
   const enriched_html = purchaseOrderData.enriched_html;
 
--- a/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
+++ b/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
@@ -1,0 +1,82 @@
+import transformEnrichment from '../transformEnrichment';
+import isEqual from 'lodash/isEqual';
+import contractData from 'components/CIDocument/components/CIDocument/__fixtures__/contract.json';
+import invoiceData from 'components/CIDocument/components/CIDocument/__fixtures__/invoice-index_op.json';
+import purchaseOrderData from 'components/CIDocument/components/CIDocument/__fixtures__/po-index_op.json';
+import { EnrichedHtml } from '../../../components/CIDocument/types';
+
+describe('transformEnrichment', () => {
+  let trasnformedDoc: EnrichedHtml[];
+  const enriched_html = contractData.enriched_html;
+
+  beforeAll(() => {
+    // trasnform doc for use in tests
+    trasnformedDoc = transformEnrichment(enriched_html);
+  });
+
+  it('takes an array, and returns an array', () => {
+    expect(Array.isArray(enriched_html)).toBeTruthy();
+    expect(Array.isArray(trasnformedDoc)).toBeTruthy();
+  });
+
+  it('adds metadata to contracts', () => {
+    expect(isEqual(enriched_html, trasnformedDoc)).toBeFalsy();
+    expect(trasnformedDoc[0].contract.metadata).toHaveLength(2);
+    expect(trasnformedDoc[0].contract.metadata[0].metadataType).toEqual('effective_dates');
+  });
+
+  it('does not add attributes and relations to contracts', () => {
+    expect(trasnformedDoc[0].contract.attributes).toBeUndefined();
+    expect(trasnformedDoc[0].contract.relations).toBeUndefined();
+  });
+});
+
+describe('transformEnrichment', () => {
+  let trasnformedDoc: EnrichedHtml[];
+  const enriched_html = invoiceData.enriched_html;
+
+  beforeAll(() => {
+    // parse doc for use in tests
+    trasnformedDoc = transformEnrichment(enriched_html);
+  });
+
+  it('adds attributes and relations to invoices', () => {
+    expect(isEqual(enriched_html, trasnformedDoc)).toBeFalsy();
+    expect(trasnformedDoc[0].invoice.attributes).toHaveLength(5);
+    expect(trasnformedDoc[0].invoice.attributes[0]).toEqual({
+      type: 'suppliers',
+      text: 'INVOICE BNP Paribas S.A.',
+      location: { end: 8831, begin: 8807 }
+    });
+    expect(trasnformedDoc[0].invoice.relations).toHaveLength(6);
+  });
+
+  it('does not add metadata to invoices', () => {
+    expect(trasnformedDoc[0].invoice.metadata).toBeUndefined();
+  });
+});
+
+describe('transformEnrichment', () => {
+  let trasnformedDoc: EnrichedHtml[];
+  const enriched_html = purchaseOrderData.enriched_html;
+
+  beforeAll(() => {
+    // parse doc for use in tests
+    trasnformedDoc = transformEnrichment(enriched_html);
+  });
+
+  it('adds attributes and relations to purchase order data', () => {
+    expect(isEqual(enriched_html, trasnformedDoc)).toBeFalsy();
+    expect(trasnformedDoc[0].purchase_order.attributes).toHaveLength(5);
+    expect(trasnformedDoc[0].purchase_order.attributes[0]).toEqual({
+      type: 'purchase_order_numbers',
+      text: '4620257561',
+      location: { end: 8319, begin: 8309 }
+    });
+    expect(trasnformedDoc[0].purchase_order.relations).toHaveLength(5);
+  });
+
+  it('does not add metadata to purchase orders', () => {
+    expect(trasnformedDoc[0].purchase_order.metadata).toBeUndefined();
+  });
+});
--- a/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
+++ b/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
@@ -1,5 +1,4 @@
 import transformEnrichment from '../transformEnrichment';
-import isEqual from 'lodash/isEqual';
 import contractData from 'components/CIDocument/components/CIDocument/__fixtures__/contract.json';
 import invoiceData from 'components/CIDocument/components/CIDocument/__fixtures__/invoice-index_op.json';
 import purchaseOrderData from 'components/CIDocument/components/CIDocument/__fixtures__/po-index_op.json';
@@ -20,7 +19,6 @@ describe('transformEnrichment - contracts', () => {
   });
 
   it('adds metadata to contracts', () => {
-    expect(isEqual(enriched_html, transformedDoc)).toBeFalsy();
     expect(transformedDoc[0].contract.metadata).toHaveLength(2);
     expect(transformedDoc[0].contract.metadata[0].metadataType).toEqual('effective_dates');
   });
@@ -41,7 +39,6 @@ describe('transformEnrichment - Invoice', () => {
   });
 
   it('adds attributes and relations to invoices', () => {
-    expect(isEqual(enriched_html, transformedDoc)).toBeFalsy();
     expect(transformedDoc[0].invoice.attributes).toHaveLength(5);
     expect(transformedDoc[0].invoice.attributes[0]).toEqual({
       type: 'suppliers',
@@ -66,7 +63,6 @@ describe('transformEnrichment - Purchase orders', () => {
   });
 
   it('adds attributes and relations to purchase order data', () => {
-    expect(isEqual(enriched_html, transformedDoc)).toBeFalsy();
     expect(transformedDoc[0].purchase_order.attributes).toHaveLength(5);
     expect(transformedDoc[0].purchase_order.attributes[0]).toEqual({
       type: 'purchase_order_numbers',
--- a/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
+++ b/packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
@@ -3,80 +3,80 @@ import isEqual from 'lodash/isEqual';
 import contractData from 'components/CIDocument/components/CIDocument/__fixtures__/contract.json';
 import invoiceData from 'components/CIDocument/components/CIDocument/__fixtures__/invoice-index_op.json';
 import purchaseOrderData from 'components/CIDocument/components/CIDocument/__fixtures__/po-index_op.json';
-import { EnrichedHtml } from '../../../components/CIDocument/types';
+import { EnrichedHtml } from 'components/CIDocument/types';
 
 describe('transformEnrichment', () => {
-  let trasnformedDoc: EnrichedHtml[];
+  let transformedDoc: EnrichedHtml[];
   const enriched_html = contractData.enriched_html;
 
   beforeAll(() => {
     // trasnform doc for use in tests
-    trasnformedDoc = transformEnrichment(enriched_html);
+    transformedDoc = transformEnrichment(enriched_html);
   });
 
   it('takes an array, and returns an array', () => {
     expect(Array.isArray(enriched_html)).toBeTruthy();
-    expect(Array.isArray(trasnformedDoc)).toBeTruthy();
+    expect(Array.isArray(transformedDoc)).toBeTruthy();
   });
 
   it('adds metadata to contracts', () => {
-    expect(isEqual(enriched_html, trasnformedDoc)).toBeFalsy();
-    expect(trasnformedDoc[0].contract.metadata).toHaveLength(2);
-    expect(trasnformedDoc[0].contract.metadata[0].metadataType).toEqual('effective_dates');
+    expect(isEqual(enriched_html, transformedDoc)).toBeFalsy();
+    expect(transformedDoc[0].contract.metadata).toHaveLength(2);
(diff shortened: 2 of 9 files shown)


packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
```diff
@@ -0,0 +1,82 @@
+import transformEnrichment from '../transformEnrichment';
+import isEqual from 'lodash/isEqual';
+import contractData from 'components/CIDocument/components/CIDocument/__fixtures__/contract.json';
+import invoiceData from 'components/CIDocument/components/CIDocument/__fixtures__/invoice-index_op.json';
+import purchaseOrderData from 'components/CIDocument/components/CIDocument/__fixtures__/po-index_op.json';
+import { EnrichedHtml } from '../../../components/CIDocument/types';
```
maniax89: this can just be
```suggestion
import { EnrichedHtml } from 'components/CIDocument/types';
```

packages/discovery-react-components/src/components/CIDocument/types.ts
```diff
@@ -46,7 +46,11 @@ export interface ItemMap {
   bySection: any;
 }
 
-export type EnrichedHtml = EnrichedHtmlContract | EnrichedHtmlInvoice | EnrichedHtmlPurchaseOrder;
+export type EnrichedHtml =
+  | EnrichedHtmlContract
+  | EnrichedHtmlInvoice
+  | EnrichedHtmlPurchaseOrder
+  | any;
```
maniax89: adding `any` kind of defeats the purpose of having the type at all, no?

packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+    expect(trasnformedDoc[0].invoice.relations).toHaveLength(6);
+  });
+
+  it('does not add metadata to invoices', () => {
+    expect(trasnformedDoc[0].invoice.metadata).toBeUndefined();
+  });
+});
+
+describe('transformEnrichment', () => {
+  let trasnformedDoc: EnrichedHtml[];
```
maniax89: ~trasnformedDoc~ transformedDoc

packages/discovery-react-components/src/components/CIDocument/types.ts
anibapat: So the EnrichedHtml can have other enrichments as well other than `contract`, `invoices`, `purchaseOrder`. eg. `tables` etc. I am not really sure what would all the other values would be so had to put any

packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
anibapat: For some reason, VSCode **was** giving error for the other import, now it is not 😄

packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
```diff
@@ -0,0 +1,82 @@
+import transformEnrichment from '../transformEnrichment';
+import isEqual from 'lodash/isEqual';
+import contractData from 'components/CIDocument/components/CIDocument/__fixtures__/contract.json';
+import invoiceData from 'components/CIDocument/components/CIDocument/__fixtures__/invoice-index_op.json';
+import purchaseOrderData from 'components/CIDocument/components/CIDocument/__fixtures__/po-index_op.json';
+import { EnrichedHtml } from 'components/CIDocument/types';
+
+describe('transformEnrichment', () => {
```
Steven-Gassert: having this describe block mention that it's used for contract data would make this a bit more readable to me, since there are several `transformEnrichment` blocks. Maybe you could tweak the other blocks as well to mention the type of data the test's relate to

packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+  let transformedDoc: EnrichedHtml[];
+  const enriched_html = contractData.enriched_html;
+
+  beforeAll(() => {
+    // trasnform doc for use in tests
+    transformedDoc = transformEnrichment(enriched_html);
+  });
+
+  it('takes an array, and returns an array', () => {
+    expect(Array.isArray(enriched_html)).toBeTruthy();
```
Steven-Gassert: hopefully we can do away with checks like these once we are able to type check our JSON blobs. I believe the work being done [here](https://github.ibm.com/Watson-Discovery/disco-widgets/issues/468) would also relate to our JSON blobs used in unit testing

anibapat: This issue was about the returning type, so I have added this test explicitly. I think it is good to have this extra layer.

packages/discovery-react-components/src/components/CIDocument/types.ts
jhpedemonte: If we don't know what the other values will be, then we won't be referencing those anywhere else. Therefore, having explicit types (not `any`) will be fine from a typescript point-of-view.

If we are referencing `tables` somewhere else, then that should be part of this type as well.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -146,14 +147,11 @@ function getAllAttributesInRelation({ relations }: Relations): Attributes[] {
...
-  if (attr && attr.location && attr.location.begin && attr.location.end) {
-    return true;
-  }
-  return false;
+  return !!(attr && attr.location && attr.location.begin && attr.location.end);
 }
 
 export default function transformEnrichment(enrichedHtml: EnrichedHtml[]): EnrichedHtml[] {
-  const newEnrichedHtml = { ...enrichedHtml };
+  const newEnrichedHtml = cloneDeep(enrichedHtml);
```
jhpedemonte: Why do we need to clone?

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -146,14 +147,11 @@ function getAllAttributesInRelation({ relations }: Relations): Attributes[] {
 // Function checks if attribute is valid and has location data in it.
 // Since have observed noisy data before so this function is required.
 function isValidAttribute(attr: Attributes): boolean {
-  if (attr && attr.location && attr.location.begin && attr.location.end) {
-    return true;
-  }
-  return false;
+  return !!(attr && attr.location && attr.location.begin && attr.location.end);
```
jhpedemonte: This would be better named as `hasLocationData`

packages/discovery-react-components/src/utils/document/__tests__/transformEnrichment.spec.tsx
jhpedemonte: Actually, this does bring up a good point: the return type was already an array (`EnrichedHtml[]`), yet we were returning an object. But TS didn't call it out.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: In order to maintain immutability created a deep cloned copy and returning it after adding metadata/attributes/relations on top of it.

jhpedemonte: But we're already doing a `cloneDeep` in `processDoc`, before calling this function. So this is probably not necessary.

anibapat:
Ready for re-review!

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -145,23 +143,18 @@ function getAllAttributesInRelation({ relations }: Relations): Attributes[] {
...
 }
 
 export default function transformEnrichment(enrichedHtml: EnrichedHtml[]): EnrichedHtml[] {
-  const newEnrichedHtml = { ...enrichedHtml };
-
-  if (newEnrichedHtml && newEnrichedHtml[0]) {
-    const enrichmentName = getEnrichmentName(newEnrichedHtml[0]);
-    const enrichment = newEnrichedHtml[0][enrichmentName];
+  if (enrichedHtml && enrichedHtml[0]) {
```
noah-eigenfeld: If you're giving `enrichedHtml` the type `EnrichedHtml[]`, then do you still need to check if it is defined? I'd think that typescript would throw an error for anyone using this method if they tried to pass in a variable that could be `undefined` or `EnrichedHtml[]`, and if they ignore that error and this fails, that's their fault.

anibapat: I kept it like this for additional failsafe.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -74,9 +74,7 @@ function createAttributeObjects(
         (att: Attributes): Attributes => createAttributeObject(att, attr)
       );
     } else {
-      return (
-        isValidAttribute(parsedObject[attr]) && createAttributeObject(parsedObject[attr], attr)
-      );
+      return hasLocationData(parsedObject[attr]) && createAttributeObject(parsedObject[attr], attr);
```
jhpedemonte: This is not easy to read. If I understand this correctly, then just rewrite like this:

```javascript
} else if (hasLocationData(parsedObject[attr])) {
   createAttributeObject(parsedObject[attr], attr);
}
```

Also, `parsedObject[attr]` is used 4 times. Refactor into a local variable.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -145,23 +143,18 @@ function getAllAttributesInRelation({ relations }: Relations): Attributes[] {
 
 // Function checks if attribute is valid and has location data in it.
 // Since have observed noisy data before so this function is required.
-function isValidAttribute(attr: Attributes): boolean {
-  if (attr && attr.location && attr.location.begin && attr.location.end) {
-    return true;
-  }
-  return false;
+function hasLocationData(attr: Attributes): boolean {
+  return !!(attr && attr.location && attr.location.begin && attr.location.end);
```
jhpedemonte: Should be able to do optional chaining now:

`return !!(attr?.location?.begin && attr?.location?.end)`

However, that still won't always work, because `begin` or `end` (well, probably only `begin`) could be zero, which would be seen as **falsy**. Probably better to check that they are not `undefined` using `typeof`.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -145,23 +143,18 @@ function getAllAttributesInRelation({ relations }: Relations): Attributes[] {
...
+  if (enrichedHtml && enrichedHtml[0]) {
+    const enrichmentName = getEnrichmentName(enrichedHtml[0]);
+    const enrichment = enrichedHtml[0][enrichmentName];
     if (enrichment) {
       modelMapping[enrichmentName](enrichment, enrichmentName);
     }
   }
 
-  return newEnrichedHtml;
+  return enrichedHtml;
```
jhpedemonte: It's not evident what is happening here. It seems that `modelMapping` is updating something inside of `enrichment`, which is a part of `enrichedHtml`. So that would be considered a function that has side-effects, resulting in not easily readable/maintainable code.

I think a better approach would be to make the `modelMapping` functions **pure**: they take in some parameters and return a value, but do not in any way change anything with the incoming parameters.

Then, in this function, we would have something like:

```javascript
const updatedEnrichment = modelMapping[enrichmentName](enrichment, enrichmentName);
enrichedHtml[0][enrichmentName] = updatedEnrichment;
```

That way, as someone who is reading this function, I can clearly see that the function `modelMapping` is returning some value, which is then stored within `enrichedHtml`.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: We can now use **Nullish Coalescing Operator - ??** which will take care of it.

jhpedemonte: Well, `??` is more of an addition to `||`, but you're using `&&` here.

anibapat: Hmm.. for some reason VSCode is giving errors to ES2020 features. Have you configured anything for that?

jhpedemonte: Oh, I thought we supported it. Perhaps ask in Slack.

anibapat:
@jhpedemonte Ready for re-review

anibapat:
Prettier and TS updates are messing with other files, I would open another PR to update those separately.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: When TS, prettier upgrade PR goes in, will add the optional chaining in the same PR.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -6,19 +6,34 @@ import { ENRICHMENTS, getEnrichmentName } from 'components/CIDocument/utils/enri
...
 } from 'components/CIDocument/types';
 import { Ontology } from './ontology/types';
 
-const modelMapping = {
-  contract: setMetadata,
-  invoice: setAttributesAndRelations,
-  purchase_order: setAttributesAndRelations
+const ontologyModelMapping = {
+  invoice: invoiceOntology,
+  purchase_order: purchaseOrderOntology
```
jhpedemonte: It seems that our ontology mapping is missing one of our ontologies ("contract"). From a maintainability standpoint, we should try to do as much of the same thing across all of our ontologies as we can. We should special case `contracts`.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -6,19 +6,34 @@ import { ENRICHMENTS, getEnrichmentName } from 'components/CIDocument/utils/enri
...
 };
 
-function setMetadata(enrichedHtml: Contract): void {
+function getupdatedEnrichment(enrichmentName: string, enrichment: any): EnrichedHtml {
```
jhpedemonte: -> `getUpdatedEnrichment` (casing)

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -6,19 +6,34 @@ import { ENRICHMENTS, getEnrichmentName } from 'components/CIDocument/utils/enri
...
+  } else if (
+    enrichmentName === ENRICHMENTS.INVOICE ||
+    enrichmentName === ENRICHMENTS.PURCHASE_ORDER
+  ) {
+    const { attributes, relations } = setAttributesAndRelations(enrichmentName, enrichment);
+    enrichment.attributes = attributes;
+    enrichment.relations = relations;
+  }
+  return enrichment;
+}
```
jhpedemonte: For example, here, why can't we just do:

```javascript
function getUpdatedEnrichment(..) {
    if (enrichmentName in ontologyModelMapping) {
        enrichment = ontologyModelMapping[enrichmentName](enrichment);
    }
    return enrichment;
}
```

That way, all these bits that update the `enrichment` object are handled by the functions defined in the mapping object.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: Because `contract` does not have `ontology`. We introduced ontology for mapping attributes and relations.

jhpedemonte: Well, perhaps we're using different meanings!

What I mean to say is that we handle 3 different types of docs. We should strive to make our code consistent in how these 3 are treated (even if 2 of them are quite close). That's what I was suggesting with my comment below about changing the `getUpdatedEnrichment` function to work with this mapping.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -38,46 +52,37 @@ function setMetadata(enrichedHtml: Contract): void {
...
-  if (ontology) {
-    doc.attributes = createAttributeObjects(ontology.attributes, doc);
-    doc.relations = createRelationObjects(ontology, doc);
-  }
+function setAttributesAndRelations(
+  enrichment: Invoice | PurchaseOrder,
+  enrichmentName: string
+): Invoice | PurchaseOrder {
+  const ontology = ontologyMapping[enrichmentName];
+  enrichment.attributes = setAttributes(ontology.attributes, enrichment);
```
jhpedemonte: I think this would be cleaner if `setAttributes` took `ontology` directly, then within that function found the `ontology.attributes` prop.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -112,8 +117,8 @@ function createRelationObject(
 
   const result = relationArray.map(
     (rel: Relations): Relations => {
-      const attributes = createAttributeObjects(ontology.attributes, rel);
-      const relations = createRelationObjects(ontology, rel);
+      const attributes = setAttributes(ontology.attributes, rel);
```
jhpedemonte: Same here.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
jhpedemonte: This is still wrong: see my comment above after "However". `begin` can be zero, in which case this will incorrectly return `false`.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -2,23 +2,37 @@ import invoiceOntology from './ontology/invoices';
...
+
 const modelMapping = {
   contract: setMetadata,
   invoice: setAttributesAndRelations,
   purchase_order: setAttributesAndRelations
 };
 
-function setMetadata(enrichedHtml: Contract): void {
+function getUpdatedEnrichment(enrichment: EnrichedHtml, enrichmentName: string): EnrichedHtml {
+  if (enrichmentName in modelMapping) {
```
ColbyJohnIBM: I'd probably prefer something like this for this function (using `lodash.get`):

```
let mapping = get(modelMapping, [enrichmentName]);
return mapping ? mapping(enrichment, enrichmentName) : enrichment;
```

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
ColbyJohnIBM: Agreed, this would make it aligned with the setRelations function.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -38,46 +52,35 @@ function setMetadata(enrichedHtml: Contract): void {
...
-  if (ontology) {
-    doc.attributes = createAttributeObjects(ontology.attributes, doc);
-    doc.relations = createRelationObjects(ontology, doc);
-  }
+function setAttributesAndRelations(
+  enrichment: Invoice | PurchaseOrder,
+  enrichmentName: string
+): Invoice | PurchaseOrder {
+  const ontology = ontologyMapping[enrichmentName];
+  enrichment.attributes = setAttributes(ontology, enrichment);
```
ColbyJohnIBM: Ideally, we make the input objects immutable, and return a new object with any data adjustments that we need.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
```diff
@@ -145,23 +148,21 @@ function getAllAttributesInRelation({ relations }: Relations): Attributes[] {
...
-    const enrichmentName = getEnrichmentName(newEnrichedHtml[0]);
-    const enrichment = newEnrichedHtml[0][enrichmentName];
-    if (enrichment) {
-      modelMapping[enrichmentName](enrichment, enrichmentName);
-    }
+  if (enrichedHtml && enrichedHtml[0]) {
+    const enrichmentName = getEnrichmentName(enrichedHtml[0]);
+    const enrichment = enrichedHtml[0][enrichmentName];
+    const updatedEnrichment = getUpdatedEnrichment(enrichment, enrichmentName);
+    enrichedHtml[0][enrichmentName] = updatedEnrichment;
```
ColbyJohnIBM: As I noted above, we should probably make the input arguments immutable.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: We already making a cloned copy of EnrichedHtml in processDoc so not doing it again here in transformEnrichment file.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: https://github.com/watson-developer-cloud/discovery-components/pull/69/files#r377243755

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
anibapat: I think the existing one is more clear readable.

packages/discovery-react-components/src/utils/document/transformEnrichment.ts
ColbyJohnIBM: It's more about the principle of immutability that I'm concerned with. We shouldn't be making changes to input arguments or using those variables for anything that isn't read-only within the function.

anibapat: I was doing that initially, but based on the comment, changed that back https://github.com/watson-developer-cloud/discovery-components/pull/69#discussion_r370415062

ColbyJohnIBM: Ah, didn't see that. I guess I'm overruled. I really would prefer immutability, but I'll bow out to @jhpedemonte.

jhpedemonte: Yeah, immutability is a larger discussion to be had.

PR #71, opened 2020-01-23
Feat/232 add ci document cypress test

broulaye:
Add CI Document to example app and add cypress test

Run Cypress test

Yes

No

https://zenhub.ibm.com/app/workspaces/discovery-unified-5728bf39d2c20f776e05d0ef/issues/watson-discovery/docviz-squad-issue-tracker/232


examples/discovery-search-app/src/App.js
```diff
@@ -3,8 +3,9 @@ import DiscoveryV2 from 'ibm-watson/discovery/v2';
 import { NoAuthAuthenticator } from 'ibm-watson/auth';
 import './app.scss';
 import { settings } from 'carbon-components';
-import { Button } from 'carbon-components-react';
+import { Button, Tabs, Tab } from 'carbon-components-react';
 import Close from '@carbon/icons-react/lib/close/16';
+import cx from 'classnames';
```
maniax89: make sure to add `classnames` as a dependency for the example app

broulaye: Good catch

broulaye:
@jhpedemonte This is ready for a re-review

broulaye:
@maniax89 This is ready for a re-review

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -90,5 +90,27 @@ describe('Passage Results', () => {
...
+    describe('and clicking on "CI Document" for a result', () => {
+      beforeEach(() => {
+        cy.get('button[data-testid="search-result-element-preview-button"]')
+          .contains('View table in document')
+          .click();
+        cy.findByText('CI');
+        cy.findByText('CI').click();
+      });
+
+      it('verify that CI Document contains filters', () => {
```
ColbyJohnIBM: These messages should describe what you expect the application to do, not what the test should do (if that makes sense). (e.g. 'displays the filters')

broulaye: make sense

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -90,5 +90,27 @@ describe('Passage Results', () => {
       cy.get('@combinedResult').contains('View passage in document');
       cy.get('@combinedResult').contains('View table in document');
     });
+
+    describe('and clicking on "CI Document" for a result', () => {
```
jhpedemonte: There's a section above: 

`describe('and clicking on "View passage in document" for a result', () => {`

where this code should live. Between the `navigate to` and `close` tests. That is, create another nested `describe` for clicking on the CI tab and running these two tests.

examples/discovery-search-app/src/App.js
```diff
@@ -85,9 +88,20 @@ function SearchPage() {
...
 
+  const tabs = [
+    {
+      name: 'Document',
+      Component: DocumentPreview
+    }
+  ];
+
+  if (canRenderCIDocument(document)) {
+    tabs.push({ name: 'CI', Component: CIDocument });
```
jhpedemonte: I know it's long, but I think it would be better to have this as `Content Intelligence` instead of `CI`, which may not be easily recognizable by users.

broulaye:
@jhpedemonte This is ready for a re-review

examples/discovery-search-app/cypress.json
```diff
@@ -1,4 +1,5 @@
 {
   "baseUrl": "http://localhost:3000/",
-  "supportFile": "cypress/support/index.ts"
+  "supportFile": "cypress/support/index.ts",
+  "defaultCommandTimeout": 30000
```
jhpedemonte: Why is this necessary?

broulaye: Because otherwise some cypress tests timeout before the page had the time to render.

jhpedemonte: Then perhaps we need to either fix our tests or our code. I believe the default is `10000` (10s) -- that should be more than enough time. Please open an issue to fix this, so we can remove this line.

Also, @H-Fin created a smaller contract in order to cut down on how long tests take. Are we using that here?

broulaye: Probably not, since this was merged before that I believe. I can create an issue to fix this

PR #73, opened 2020-01-24
Fix: 204 faster CI Doc tests and more tests

H-Fin:
Unit tests for CI Document no longer take 20 seconds each, all are around 0.5 seconds. There are also more tests implemented to increase our code coverage.

https://zenhub.ibm.com/app/workspaces/discovery-unified-5728bf39d2c20f776e05d0ef/issues/watson-discovery/docviz-squad-issue-tracker/204

`yarn test CIDocument` and see that the tests pass.

`Run yarn test:coverage` and view the function and branch coverage of the `CIDocument\utils\enrichmentUtils.ts` is up to statements: 100%, branches: 83.33%, and functions: 100% .

N/A

No


packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -327,6 +327,7 @@ const CIDocument: FC<CIDocumentProps> = ({
             selected={nonContractTabs.indexOf(selectedType)}
             onSelectionChange={(index: number): void => setSelectedType(nonContractTabs[index])}
             aria-label={messages.filtersTabLabel}
+            data-testid="tabs"
```
ColbyJohnIBM: I think I'd prefer most descriptive / unique `testid` names. That being said, I'm bad at naming things. 😆 Maybe `noncontract-tabs` or `cidocument-noncontract-tabs`?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -350,6 +351,7 @@ const CIDocument: FC<CIDocumentProps> = ({
               onSelectionChangeContractTabs(index, setSelectedContractFilter, resetStates)
             }
             aria-label={messages.filtersTabLabel}
+            data-testid="tabs"
```
ColbyJohnIBM: Same deal here... `contract-tabs`?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/__tests__/CIDocument.spec.tsx
```diff
@@ -9,72 +9,154 @@ import {
...
-  FindByText
+  FindByText,
+  QueryByText
 } from '@testing-library/react';
 import 'utils/test/createRange.mock';
 import CIDocument from '../CIDocument';
-import contract from '../__fixtures__/contract.json';
+import purchaseOrder from '../__fixtures__/po-index_op.json';
+import invoice from '../__fixtures__/invoice-index_op.json';
+import shortContract from '../__fixtures__/shortenedContract.json';
```
ColbyJohnIBM: Curiosity question: reason for using shorted contract?

ColbyJohnIBM: Nevermind, I see the code comment below. :+1:

broulaye:
I'm curious to know what is it that was making the test take 20 sec, from what I can see you're still rendering `CIDocument` before each test

H-Fin:
> I'm curious to know what is it that was making the test take 20 sec, from what I can see you're still rendering `CIDocument` before each test

The contract document we were using to test with was large compared to the invoice and purchase order documents. This was causing a long run time for each test. I modified the contract document to create a shorter version as well as applied most tests to the other document types.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -350,6 +351,7 @@ const CIDocument: FC<CIDocumentProps> = ({
               onSelectionChangeContractTabs(index, setSelectedContractFilter, resetStates)
             }
             aria-label={messages.filtersTabLabel}
+            data-testid="contract-tabs"
```
jhpedemonte: 1. There is a lot of duplication in this `<Tabs>` code; could use refactoring to cut down on that (but that's a separate issue)
2. The first case is actually invoice/PO, not contract...
3. But it seems you've noticed that only one of these is on screen at same time and we should use same test ID, to keep things simpler.
4. I suggest something like `document-info-tabs`

packages/discovery-react-components/src/components/CIDocument/components/DetailsPane/DetailsPane.tsx
```diff
@@ -20,7 +20,7 @@ const DetailsPane: FC<DetailsPaneProps> = ({
   onActiveLinkChange
 }) => {
   return (
-    <div className={base}>
+    <div className={base} data-testid="detailsPane">
```
jhpedemonte: Let's keep convention of [kebab case](https://en.wikipedia.org/wiki/Letter_case#Special_case_styles): `details-pane`

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
H-Fin: Good point. Went ahead and fixed those.

PR #74, opened 2020-01-27
chore: add unit test for CIDocument details utility

JuanNicolasGarcia:

Adds unit tests to the CIDocument details utility for higher test coverage of our code.

`yarn test details` and see that details util tests pass.

Run `yarn test:coverage` and view the function and branch coverage of the CIDocument\utils\details.ts file as see the results are above desired threshold 80% - branch, 100% - function.


JuanNicolasGarcia:

packages/discovery-react-components/src/components/CIDocument/utils/details.ts
```diff
@@ -52,7 +52,7 @@ function getContractAttributes(attributes: any): Item[] {
   });
 }
 
-function getAttributesWithinRelation(attributes: any): Item[] {
+export function getAttributesWithinRelation(attributes: any): Item[] {
```
anibapat: why is it exported? Just so it can be tested individually in the tests?

packages/discovery-react-components/src/components/CIDocument/utils/__tests__/details.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+} from '../details';
+
+const mockData = {
+  metadataType: 'effective_dates',
+  text: '24th October 2011'
+};
+
+const mockDataNoMeta = { text: '24th October 2011' };
+const mockDetails = [{ heading: 'type', items: ['effective_dates'] }];
+const mockDetailsNoMeta = [{ heading: 'party', items: ['24TH OCTOBER 2011'] }];
```
broulaye: what's the difference between `mockDetails` and `mockDetailsNoMeta`? seems to be the same to me beside the value in it

packages/discovery-react-components/src/components/CIDocument/utils/details.ts
JuanNicolasGarcia: It is mainly for testing purposes.  I use this function to get the specific output needed to do the test compare. The alternative would be to mock the return value for that function.

packages/discovery-react-components/src/components/CIDocument/utils/__tests__/details.spec.ts
JuanNicolasGarcia: The difference between the two are test the different outputs from the function `getDetailsFromMetadata`.  The heading and items values will change based on if a metadataType is present in the data prop passed in.

JuanNicolasGarcia: Will change the variable names to be more clear.

PR #75, opened 2020-01-28
chore: add unit test for documentPreview utils

H-Fin:
Adds unit tests to the documentPreview utils file `documentData.ts` for higher test coverage of our code.

https://zenhub.ibm.com/app/workspaces/discovery-unified-5728bf39d2c20f776e05d0ef/issues/watson-discovery/docviz-squad-issue-tracker/234

`yarn test watch:documentData` and see that tests pass.

Run `yarn test:coverage` and view the coverage of the `DocumentPreview/utils/documentData.ts` file to see the results are statements: 100%, branches: 100%, and functions: 100%.


packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -15,11 +15,17 @@ export function getTextMappings(
...
         console.error('Failure parsing text_mappings', err);
         mappings = null;
       }
+    } else {
+      // text mapping is not a string
+      mappings = null;
     }
+  } else {
+    // no text mappings are present
+    mappings = null;
```
jhpedemonte: Instead of these 2 separate but identical else branches, I think it would be better to slightly refactor the code:

```
if (mappings && typeof mappings === 'string') {
    ...
} else {
    // text_mappings are not what we expect; reset to null
    mappings = null;
}
```

packages/discovery-react-components/src/components/DocumentPreview/utils/__tests__/documentData.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  afterAll(() => {
+    consoleError.mockRestore();
+  });
+
+  it('succeeds in getting text mapping from json file', () => {
+    const mappings = getTextMappings(jsonDoc);
+    expect(mappings).not.toEqual(null);
+    expect(consoleError).not.toHaveBeenCalled();
+  });
```
jhpedemonte: Good test here of using real data. But I think you should also add "good" dummy data (with `text_mappings` as a JSON string) and specifically test that the returned data is an object with the expected properties.

H-Fin: Went ahead and made those changes!

packages/discovery-react-components/src/components/DocumentPreview/utils/__tests__/documentData.test.ts
```diff
@@ -35,6 +42,16 @@ describe('documentData', () => {
     expect(consoleError).not.toHaveBeenCalled();
   });
 
+  it('returns correct text mapping data', () => {
+    const mappings = getTextMappings(correctTextMapping)!;
+    expect(mappings).not.toEqual(null);
+    expect(mappings.pages[0].page_number).toEqual(0);
+    expect(mappings.pages[0].height).toEqual(792);
+    expect(mappings.pages[0].width).toEqual(612);
+    expect(mappings.pages[0].origin).toEqual('TopLeft');
```
jhpedemonte: I think you can do a comparison against an actual object, rather than having to do each property individually.

PR #77, opened 2020-01-28
fix: added tests for idUtils to improve test coverage

anibapat:

Adds test for idUtils. 

Yarn test

No

No


packages/discovery-react-components/src/utils/document/__tests__/idUtils.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+describe('idUtils - function getId', () => {
+  it('gives back id from a valid item', () => {
+    expect(getId(elements[0])).toBe('469800_469817');
+  });
+
+  it('gives back undefined from an invalid item', () => {
+    const item: any = { text: 'Foo bar' };
+    expect(getId(item)).toBeUndefined();
+  });
```
jhpedemonte: Hmm, this points out a possible issue with current implementation. _If_ there is invalid data, there could be multiple with all the same ID of `undefined`. Could this cause any problems? Would it be better to throw an error or somehow fail gracefully?

anibapat: I have not seen such a scenario/case till now, but yes it would give the same `undefined` id, (could cause issues on the UI, as it will always go to the 1st element, etc. )

PR #78, opened 2020-01-28
Feat/add project fields

Steven-Gassert:
This fixes the bug that was previously merged due to our ESLint rules automatically adding `fetchFields` as a dependency to the useEffect hook which is in charge of calling it. This previously caused us to be in a loop of indefinitely calling `fetchFields`

- Load up storybook and open the `StructuredQuery` component. Notice that `fetchFields` only gets called once
n/a
no


Steven-Gassert:
closing this PR to make these changes elsewhere

PR #80, opened 2020-01-28
chore: add script to setup and run example app

ChrisGillin:

Add interactive script for configuring and building example application. Also prompts user asking if they'd like to start the example app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,13 @@ The example app is a catalogue of the core components provided by this library. 
 
 The `runExampleApp.sh` script (located in `examples/discovery-search-app/scripts`) provides prompts to help configure and run the example application. The script iterates through the following steps:
 
-1. Verify all prerequite programs are installed
-1. Prompt user for necessary cluster information
+1. Verify all prerequisite programs are installed
+1. Prompt you for necessary cluster information
 1. Configure the example application server
 1. Build the React components
-1. Ask the user if they'd like to start the example application
+1. Ask you if you'd like to start the example application
 
-If a user chooses not to start the example application, all previous configuration steps will be left intact so the application can be run at another time by running
+If you chooses not to start the example application, all previous configuration steps will be left intact so the application can be run at another time by running
 
 ```
 yarn workspace discovery-search-app run start
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 - [Running the example app](#running-the-example-app)
   * [Setup script](#setup-script)
   * [Manual setup](#manual-setup)
+    + [Windows Only](#windows-only)
 - [Using Discovery Components in a React application](#using-discovery-components-in-a-react-application)
   * [Interacting with Discovery data in custom components](#interacting-with-discovery-data-in-custom-components)
 - [Development](#development)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -44,6 +46,24 @@
 
 The example app is a catalogue of the core components provided by this library. With little effort, you can see the functionality of each component using your real data. You can also modify the example code to see how you can customize Discovery Components to fit your needs.
 
+### Setup script
+
+The `runExampleApp.sh` script (located in `examples/discovery-search-app/scripts`) provides prompts to help configure and run the example application. The script iterates through the following steps:
+
+1. Verify all prerequite programs are installed
```
kabeaty: `prerequite` --> `prerequisite` :)

README.md
```diff
@@ -44,6 +46,24 @@
...
+1. Prompt user for necessary cluster information
+1. Configure the example application server
+1. Build the React components
+1. Ask the user if they'd like to start the example application
+
+If a user chooses not to start the example application, all previous configuration steps will be left intact so the application can be run at another time by running
```
kabeaty: To stay consistent with the rest of the README, I think we should use `you` over `user` here and in the above updates. For example, this sentence would then be something like: `If you choose not to start the example application, all previous...`

examples/discovery-search-app/scripts/runExampleApp.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
+
+set -e
+
+PREREQUISTES=("yarn" "jq")
```
kabeaty: Can you update the spelling here and below too to `PREREQUISITES`? :)

README.md
```diff
@@ -44,6 +46,24 @@
 
 The example app is a catalogue of the core components provided by this library. With little effort, you can see the functionality of each component using your real data. You can also modify the example code to see how you can customize Discovery Components to fit your needs.
 
+### Setup script
```
noah-eigenfeld: Small nitpick: can you add the command to run the script, and maybe specify which directory to run it from?
```
./examples/discovery-search-app/scripts/runExampleApp.sh
```
or, if it's supposed to be run in the example app dir, this:
```
cd examples/discovery-search-app && ./scripts/runExampleApp.sh

File: examples/discovery-search-app/scripts/runExampleApp.sh [hunk 5]
```diff
@@ -0,0 +1,151 @@
...
+  read -sp "  password: " password
+
+  if [[ -z $host || -z $port || -z $username || -z $password ]]; then
+    echo
+    echo
+    echo "You provide all of the following: host, port, username and password" >&2
+    exit 1
+  fi
+
+  cat >"$SCRIPT_DIR/../.env.local" <<EOL
```
noah-eigenfeld: When users want to set up Discovery Components in their own applications, are they expected to populate `.env` or `.env.local`? If we want them to put their credentials in `.env` at some point, it might be a good idea to call out in the readme that this script only populates `.env.local`.

noah-eigenfeld: If we switch `.env` -> `.env.template`, that would ease confusion around this point

README.md
```diff
@@ -82,6 +103,13 @@ The example app is a catalogue of the core components provided by this library.
    4. `CLUSTER_PORT` defaults to `443` unless configured otherwise
    5. `CLUSTER_HOST` the base URL of your CP4D cluster (ex. `example.com`)
 
+    #### Windows Only
+    On Windows, the default `SASS_PATH` environmnt variable must be updated. Append the following to the `.env.local` file:
```
noah-eigenfeld: I think this (combined with the override in the setup script) is a good enough solution for setting up `SASS_PATH` across multiple environments

README.md
```diff
@@ -44,6 +47,29 @@
...
+1. Configure the example application server
+1. Build the React components
+1. Ask you if you'd like to start the example application
+
+Run the following command from the project root directory
+```
+./runExampleApp.sh
+```
+
+If you chooses not to start the example application, all previous configuration steps will be left intact so the application can be run at another time by running
```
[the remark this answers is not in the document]
ChrisGillin: Ugh...just fixed it in place.

PR #83, opened 2020-01-31
feat: structured query convert inputs

kabeaty:
These changes add in the conversion and display of the query that matches the selections made in the `StructuredQuery` component.

Note that I didn't add Cypress tests in this PR, as I think it makes sense to add those in in a separate issue.

You can take a look at and run the tests. You can also play around with the `StructuredQuery` component in Storybook, selecting fields, operators, and entering values and watching the copyable query at the bottom of the page update. Confirm that the query looks as you would expect it to look.

not in this PR, there is another issue to update the documentation for the `StructuredQuery` component

no


packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
```diff
@@ -40,6 +55,18 @@ const StructuredQuery: FC<StructuredQueryProps> = ({ messages = defaultMessages
...
 
+  const handleCopyQueryOnClick = () => {
+    const stringifiedSelectionText = document.querySelector(
+      `.${structuredQueryCopyableQueryClass}`
+    )!.textContent;
+    const textArea = document.createElement('textarea');
+    textArea.textContent = stringifiedSelectionText;
+    document.body.appendChild(textArea);
+    textArea.select();
+    document.execCommand('copy');
```
Steven-Gassert: Apparently the `execCommand` function isn't available in Safari https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard (scroll down a bit or search for the text 'safari'). Maybe we can spin up a tech_debt issue to support this?

packages/discovery-react-components/src/components/StructuredQuery/components/RemoveRuleRowButton/RemoveRuleRowButton.tsx
```diff
@@ -36,15 +36,20 @@ export const RemoveRuleRowButton: FC<RemoveRuleRowButtonProps> = ({
...
     if (isLastRuleInRuleGroup) {
       setStructuredQuerySelection({
+        ...structuredQuerySelection,
         groups: {
           ...omit(structuredQuerySelection.groups, groupId)
         },
+        rows: {
+          ...filteredRows
+        },
         group_order: structuredQuerySelection.group_order.filter(id => id !== groupId)
```
Steven-Gassert: do you mind clarifying what the use of `group_order` is?

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
kabeaty: Ah, good catch. Yeah, I can do that 👍

packages/discovery-react-components/src/components/StructuredQuery/components/RemoveRuleRowButton/RemoveRuleRowButton.tsx
kabeaty: Sure! See this discussion for the background: https://github.com/watson-developer-cloud/discovery-components/pull/42#discussion_r364437642. I can also add a comment clarifying if you think that would be helpful?

packages/discovery-react-components/src/components/StructuredQuery/StructuredQuery.tsx
kabeaty: As discussed, I pushed up a change to not include copy functionality (at least until we have clarity on the desired interaction and if it does include this)

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
```diff
@@ -0,0 +1,32 @@
+import { StructuredQuerySelection } from './structuredQueryInterfaces';
+
+export const stringifyStructuredQuerySelection = (
+  structuredQuerySelection: StructuredQuerySelection
+) => {
+  let stringifiedStructuredQuerySelection: string = '';
+  stringifiedStructuredQuerySelection += Object.keys(structuredQuerySelection.groups)
```
maniax89: i thought that `group_order` was going to be used here to grab the groups instead of doing `Object.keys` on the groups object?

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
```diff
@@ -0,0 +1,32 @@
...
+    .map(groupId => {
+      return groupId === '0'
```
maniax89: should the condition be whether or not there are multiple rows (i.e. `rows.length > 1`) before wrapping in parentheses?

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
```diff
@@ -0,0 +1,32 @@
...
+    })
+    .join(structuredQuerySelection.groups[0].operator);
+  return stringifiedStructuredQuerySelection;
+};
+
+const stringifyRows = (structuredQuerySelection: StructuredQuerySelection, groupId: string) => {
+  return structuredQuerySelection.groups[groupId].rows
+    .map((rowId: string) => {
+      const row = structuredQuerySelection.rows[rowId];
+      const rowField = containsReservedCharacters(row.field) ? '"' + row.field + '"' : row.field;
```
maniax89: i don't think we want to wrap field names in quotes as that query wouldn't actually work. is there a way we are testing the queries that this component produces?

maniax89: i believe if you want to escape the field names, you can use a backslash `\` and the query parser will still work

for example:
```
my,field:bar
```

would become
```
my\,field:bar
```

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
```diff
@@ -0,0 +1,32 @@
...
+      const rowValue = containsReservedCharacters(row.value) ? '"' + row.value + '"' : row.value;
+      return `${rowField}${row.operator}${rowValue}`;
+    })
+    .join(structuredQuerySelection.groups[groupId].operator);
+};
+
+const containsReservedCharacters = (fieldOrValue: string) => {
+  // Reserved characters include , | : ! as they are used as operators in the query
+  const regexForReservedCharacters = /([,|:!])+/g;
```
maniax89: i think there are more than these for reserved characters -> https://github.ibm.com/Watson-Discovery/discovery-query-parser/blob/master/src/main/antlr/DiscoveryLexer.g4

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
kabeaty: Ah, yep, updated...thanks for keeping me honest!

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
kabeaty: I don't think so because, at least on cloud, the initial top-level rules are never wrapped in parentheses, but then the rule(s) for all of the nested rule groups are.

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
kabeaty: Ah, okay, thanks for the catch and links with more info on this. Updated to escape the expanded set of reserved characters with a backslash 👍

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
```diff
@@ -4,29 +4,31 @@ export const stringifyStructuredQuerySelection = (
...
-const stringifyRows = (structuredQuerySelection: StructuredQuerySelection, groupId: string) => {
+const stringifyRows = (structuredQuerySelection: StructuredQuerySelection, groupId: number) => {
   return structuredQuerySelection.groups[groupId].rows
-    .map((rowId: string) => {
+    .map((rowId: number) => {
       const row = structuredQuerySelection.rows[rowId];
-      const rowField = containsReservedCharacters(row.field) ? '"' + row.field + '"' : row.field;
-      const rowValue = containsReservedCharacters(row.value) ? '"' + row.value + '"' : row.value;
+      const rowField = escapeReservedCharacters(row.field);
+      const rowValue = row.value === '' ? row.value : '"' + row.value + '"';
```
maniax89: Should we always be quoting values? I would think that we would only quote things that have reserved characters. Also we need to make sure we escape double quotes

kabeaty: Yeah, that's what I was thinking too. But I updated to always quote values since I realized that's what we're doing on cloud and that way we don't spend time checking for reserved characters when the values can just always be quoted.

I have double quotes in the regex, or maybe I'm missing something?

packages/discovery-react-components/src/components/StructuredQuery/utils/stringifyStructuredQuerySelection.ts
```diff
@@ -19,12 +19,16 @@ const stringifyRows = (structuredQuerySelection: StructuredQuerySelection, group
...
       const rowField = escapeReservedCharacters(row.field);
-      const rowValue = row.value === '' ? row.value : '"' + row.value + '"';
+      const rowValue = row.value === '' ? row.value : '"' + escapeDoubleQuotes(row.value) + '"';
       return `${rowField}${row.operator}${rowValue}`;
     })
     .join(structuredQuerySelection.groups[groupId].operator);
 };
 
+const escapeDoubleQuotes = (value: string) => {
+  return value.replace('"', '\\"');
```
maniax89: this will only replace the first occurrence of a double quote. if you use a regex with the `g` modifier (global), it will replace all matches

PR #86, opened 2020-02-03
fix: show attribute details when click on clickable attribute within element

anibapat:
Capability to show attribute details when attribute is clicked within element.

Duplicate of closed PR: https://github.ibm.com/Watson-Discovery/disco-widgets/pull/349
implements: https://github.ibm.com/Watson-Discovery/docviz-squad-issue-tracker/issues/145

`yarn storybook` -> CIDocument -> click on any attribute within element and should see details of the attribute.

No

No

### DIFF
--- a/packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
+++ b/packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
@@ -633,7 +633,7 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
 function getActiveElement(activeIds: string[], itemList: any[]): any {
   return itemList.find(item => {
     if ('allAttributeIds' in item) {
-      return isEqualId(item.allAttributeIds, activeIds);
+      return isEqual(item.allAttributeIds.sort(), activeIds.sort());
     } else if (isEqualId(item, activeIds)) {
       return true;
     } else if ('attributes' in item) {
--- a/packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
+++ b/packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
@@ -193,6 +193,7 @@ const CIDocument: FC<CIDocumentProps> = ({
     setHighlightedList([]);
     setCurrentFilter({});
     setActiveMetadataIds([]);
+    setClickedItemType('');
   };
 
   useEffect(() => {
@@ -256,6 +257,7 @@ const CIDocument: FC<CIDocumentProps> = ({
         const { filteredList, filterGroups: groups } = filterHelper.processFilter(currentFilter);
         setHighlightedList(filteredList);
         setFilterGroups(groups);
+        setClickedItemType('');
       }
     }
   }, [filterHelper, currentFilter, itemList, didCatch]);
@@ -269,6 +271,7 @@ const CIDocument: FC<CIDocumentProps> = ({
       if (hasRelation(highlightedList)) {
         setActiveIds(highlightedList[0].allAttributeIds);
       } else {
+        setClickedItemType('');
         setActiveIds([getId(highlightedList[0])]);
       }
     }
@@ -298,6 +301,14 @@ const CIDocument: FC<CIDocumentProps> = ({
     if (activeMetadataIds.length > 0 && !(activeIds && activeIds.length > 0)) {
       activeDetails = getDetailsFromMetadata(highlightedList[activeIndex]);
     }
+  }
+
+  // In contracts, only elements should be highlighted and clickable.
+  // Currently attributes can be hovered and clicked.
+  // This can be removed once hover/click ability on attributes in sections is removed.
+  const [clickedItemType, setClickedItemType] = useState<string>('');
+  if (clickedItemType === ATTRIBUTES && !isInvoiceOrPurchaseOrder(enrichmentName)) {
+    activeDetails = getDetailsForAttribute(activeElement, activeIds);
   }
 
   let highlightedIds = [];
@@ -417,7 +428,8 @@ const CIDocument: FC<CIDocumentProps> = ({
               onChange={onNavigationChange({
                 setActiveIds:
                   selectedContractFilter === METADATA ? setActiveMetadataIds : setActiveIds,
-                highlightedList
+                highlightedList,
+                setClickedItemType
               })}
             />
             <div className={`${base}__rightGutter`} />
@@ -441,7 +453,8 @@ const CIDocument: FC<CIDocumentProps> = ({
               activePartIds={activePartIds}
               onItemClick={onItemClick({
                 setActiveIds,
-                elementList: itemList
+                elementList: itemList,
+                setClickedItemType
               })}
               activeMetadataIds={activeMetadataIds}
               theme={theme}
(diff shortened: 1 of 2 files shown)


packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -611,13 +632,21 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
...
   return itemList.find(item => {
-    if (item.allAttributeIds) {
+    if ('allAttributeIds' in item) {
       return isEqual(item.allAttributeIds.sort(), activeIds.sort());
+    } else if (isEqualId(item, activeIds)) {
+      return true;
+    } else if ('attributes' in item) {
+      return item.attributes.find((attribute: any) => isEqualId(attribute, activeIds));
     }
     return isEqual([getId(item)].sort(), activeIds.sort());
```
broulaye: why not use the `isEqualid` function here

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -611,13 +632,21 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
 
 function getActiveElement(activeIds: string[], itemList: any[]): any {
```
broulaye: the return type of this function is `boolean` right? not `any`

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -611,13 +632,21 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
...
   return itemList.find(item => {
-    if (item.allAttributeIds) {
+    if ('allAttributeIds' in item) {
       return isEqual(item.allAttributeIds.sort(), activeIds.sort());
+    } else if (isEqualId(item, activeIds)) {
+      return true;
+    } else if ('attributes' in item) {
+      return item.attributes.find((attribute: any) => isEqualId(attribute, activeIds));
```
broulaye: since all of these `else-ifs` have return statements, I think it would be more readable to just separate them in their own `if statements instead` of chaining them like above, or maybe use a switch statement. This is just a personal preference of mine, if you want to keep it this way that's fine too

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
anibapat: No, it gives details about the active element, not boolean.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
anibapat: Yah I prefer if- elseIf chaining pattern.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -269,6 +271,7 @@ const CIDocument: FC<CIDocumentProps> = ({
       if (hasRelation(highlightedList)) {
         setActiveIds(highlightedList[0].allAttributeIds);
       } else {
+        setClickedItemType('');
```
jhpedemonte: I'm concerned here that we had to add `setClickedItemType('')` in multiple places. It seems our code isn't structured well if that was required.

What is the actual intent here? Are we trying to reset something while also setting something else? If so, could we do something like:

```
resetAll();
setThatOneThing();
```

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -611,13 +632,21 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
...
-    if (item.allAttributeIds) {
+    if ('allAttributeIds' in item) {
       return isEqual(item.allAttributeIds.sort(), activeIds.sort());
+    } else if (isEqualId(item, activeIds)) {
+      return true;
+    } else if ('attributes' in item) {
+      return item.attributes.find((attribute: any) => isEqualId(attribute, activeIds));
     }
-    return isEqual([getId(item)].sort(), activeIds.sort());
+    return false;
```
jhpedemonte: This should be `null`, right? `false` doesn't make sense.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -611,13 +632,21 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
...
   });
 }
 
+function isEqualId(item: any, id: string[]): boolean {
```
jhpedemonte: Try to avoid adding `any`. In this case, it seems like `item` should be an object with an `id` property, right?

jhpedemonte: Also, `id` is an array, meaning multiple. Therefore, prefer `ids` for the name.

And doesn't that mean that this should be named `isEqualIds`?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -611,13 +632,21 @@ function getActiveIndex(activeIds: string[], highlightedList: any[]): number {
...
+  return isEqual([getId(item)].sort(), id.sort());
```
jhpedemonte: This seems odd: does `getId` return a single ID? If so, why call `sort()` on a singleton array?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
jhpedemonte: Can we get a proper type for `attribute` instead of `any`?

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -626,6 +655,20 @@ function getAllClickableIds(itemList: any[]): string[] {
...
+    const attribute = activeElement.attributes.find((attr: any) => isEqualId(attr, activeIds));
+    return [
+      {
+        heading: ATTRIBUTES,
+        items: [attribute.type]
+      }
+    ];
+  }
+  return [];
+}
```
jhpedemonte: Same in this function: avoid `any` if you can.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
```diff
@@ -300,6 +303,14 @@ const CIDocument: FC<CIDocumentProps> = ({
     }
   }
 
+  // In contracts, only elements should be highlighted and clickable.
+  // Currently attributes can be hovered and clicked.
+  // This can be removed once hover/click ability on attributes in sections is removed.
+  const [clickedItemType, setClickedItemType] = useState<string>('');
```
ColbyJohnIBM: What's the difference between this and selectedType?

ColbyJohnIBM:
I think we need to re-think this a bit more-- ideally we don't have a difference between the "clicked" item and the "selected" item. I'm not sure how we should handle selecting an attribute within an element in a contract. Trying to think it through...

anibapat:
@jhpedemonte 
>Can we instead remove the ability to click on attributes and only allow clicking on elements? What would that take?

We need to investigate that. I personally like the way it is in CNC, where we do not give an ability to click on individual attributes within elements. I think that is a bit confusing too considering the user experience. 

Closing this PR and making a task for investigation.

PR #87, opened 2020-02-04
fix: better handling of tables in processDoc

H-Fin:
ProcessDoc is now able to add top level tables to the tables list.

https://zenhub.ibm.com/app/workspaces/discovery-unified-5728bf39d2c20f776e05d0ef/issues/watson-discovery/docviz-squad-issue-tracker/252

Run `yarn test:watch processDoc` and see that the newly added test passes. If you were to run the test on develop it would not.


packages/discovery-react-components/src/utils/document/__tests__/processDoc.spec.tsx
```diff
@@ -159,4 +159,42 @@ describe('processDoc', () => {
...
+
+    const tableOne = getByTestId('table-one');
+    const tableTwo = getByTestId('table-two');
+    const tableThree = getByTestId('table-three');
+    expect(tableOne.textContent).toEqual('Table 1');
+    expect(tableTwo.textContent).toEqual('Table 2');
+    expect(tableThree.textContent).toContain('Column 1' && 'Column 2');
+
+    expect(doc.tables!.length).toEqual(3);
+    expect(doc.tables![2].bboxes.length).toEqual(2);
```
jhpedemonte: I think it would be good to have a check for actual table data, against one of the entries. For example, we expect that the 2nd table will have `location` data of ...

packages/discovery-react-components/src/utils/document/__tests__/processDoc.spec.tsx
```diff
@@ -159,4 +159,42 @@ describe('processDoc', () => {
...
+    const doc = await processDoc({ html }, { sections: true, tables: true });
+    const processedHtml = doc.sections!.map(section => section.html).join('');
+    const { getByTestId } = render(<div dangerouslySetInnerHTML={{ __html: processedHtml }} />);
+
+    const tableOne = getByTestId('table-one');
+    const tableTwo = getByTestId('table-two');
+    const tableThree = getByTestId('table-three');
+    expect(tableOne.textContent).toEqual('Table 1');
+    expect(tableTwo.textContent).toEqual('Table 2');
+    expect(tableThree.textContent).toContain('Column 1' && 'Column 2');
```
jhpedemonte: I don't think this does what you think it does. `'Column 1' && 'Column 2'` will be evaluated first, becoming `true`.

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -186,14 +186,7 @@ function setupBodyParser(parser: SaxParser, doc: ProcessedDoc): void {
...
       if (SECTION_NAMES.includes(tagName)) {
-        setupSectionParser(
-          parser,
-          doc,
-          tagName,
-          attributes,
-          p.startIndex,
-          getChildBeginFromOpenTag(p)
-        );
+        setupSectionParser(parser, doc, tagName, attributes, getChildBeginFromOpenTag(p), p);
```
jhpedemonte: I think we actually want to pass `p.startIndex` instead of `getChildBeginFromOpenTag(p)` -- the latter is handled within `onopentag`.

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -322,8 +315,16 @@ function setupSectionParser(
           parser.popState();
         }
       }
+      if (tagName === 'body') {
```
jhpedemonte: Add a empty line above this. Breaking up the code can make it easier to visually parse/read.

packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -322,8 +315,16 @@ function setupSectionParser(
...
+        parser.popState();
+      }
     }
-  });
+  };
+
+  // Ensures the current section open tag is called first and then
+  // continues through tags inside of that section.
+  parser.pushState(actionState);
+  actionState.onopentag(sectionParser, sectionTagName, sectionTagAttrs);
```
jhpedemonte: One thing I mentioned to you yesterday was that I didn't like having this bit of setup code separated from the rest by the large chunk of actionState` code. It would be very easy for someone reading this code to miss these 2 lines.

Instead, I suggest that we use function hoisting to our advantage, in order to keep the setup code together:

```javascript
  // track nested nodes of same tag name
  let stackCount = 1;
  let stackCount = 0;
  const openTagIndices = [0];
  const openTagIndices = [0];
  const sectionHtml = [openTagToString(sectionTagName, sectionTagAttrs, sectionChildBegin)];
  const sectionHtml: string[] = [];

  const actionState = getActionState();
  parser.pushSTate(actionState);

  // Ensures the current section open tag is called first and then
  // continues through tags inside of that section.
  actionState.onopentag(sectionParser, sectionTagName, sectionTagAttrs);

  function getActionState() {
    return {
      onopentag: ...
    }
  }
```

H-Fin:
@jhpedemonte This is ready for a re-review 😄

PR #88, opened 2020-02-04
chore: add unit tests for elementFromPoint

JuanNicolasGarcia:

Adds unit tests for the `elementFromPoint` utility.

In discovery-react-components run:
`yarn test elementFromPoint` and see that tests pass.


packages/discovery-react-components/src/components/CIDocument/utils/__tests__/elementFromPoint.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+const mockElements = [
+  { classList: classList(['mockName']), style: { pointerEvents: null } },
+  { classList: classList(['mockClass']), style: { pointerEvents: null } },
+  { classList: classList([className]), style: { pointerEvents: null } }
+];
+
+describe('elementFromPoint', () => {
+  it('elementsFromPoint', () => {
```
jhpedemonte: Nit: Think of `describe` and `it` as parts of a sentence: 

```javascript
describe('CIDocument`, () => {
  it('renders properly with a contract', ...)
```

-> "CIDOcument renders properly with a contract".

You setup here doesn't work like that. Please fix.

packages/discovery-react-components/src/components/CIDocument/utils/__tests__/elementFromPoint.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    const originalElementsFromPoint = document.elementsFromPoint;
+    document.elementsFromPoint = (x: number, y: number): Element[] => {
+      return (mockElements as unknown) as Element[];
+    };
+
+    const elementResult = elementFromPoint(clientX, clientY, className);
+    expect(elementResult).toEqual(mockElements[2]);
+
+    document.elementsFromPoint = originalElementsFromPoint;
```
jhpedemonte: It doesn't impact things with the current tests, but best practices here would be to put the setup and cleanup in `before`/`after` statements. Otherwise, if `expect` throws an error, the cleanup is never called and tests that follow _could_ fail because of this.

So in this case, it probably makes the most sense to have a `beforeAll` that does the setup for **all** of the functions you want to override. Then have cleanup for same in `afterAll`.

packages/discovery-react-components/src/components/CIDocument/utils/__tests__/elementFromPoint.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    contains: (name: string): boolean => {
+      return list.includes(name);
+    }
+  };
+};
+
+const mockElements = [
+  { classList: classList(['mockName']), style: { pointerEvents: null } },
+  { classList: classList(['mockClass']), style: { pointerEvents: null } },
+  { classList: classList([className]), style: { pointerEvents: null } }
```
jhpedemonte: Reading this again, I think I would prefer if there were an actual string here instead of the `className` variable. Then in each test, we would use same string. Makes it more readable.

packages/discovery-react-components/src/components/CIDocument/utils/__tests__/elementFromPoint.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const elementResult = elementFromPointMs(clientX, clientY, className);
+    expect(elementResult).toEqual(mockElements[2]);
+
+    document.elementsFromPoint = originalElementsFromPoint;
+  });
+
+  it('elementFromPointFallback', () => {
+    const originalElementFromPoint = document.elementFromPoint;
+    let test = 0;
```
jhpedemonte: `test` doesn't make sense as a var name. I prefer something more descriptive like `nextElementIndex`.

JuanNicolasGarcia:
@jhpedemonte This is ready for re-review.

PR #89, opened 2020-02-05
fix: reset pagination when new query is submitted

ChrisGillin:
When a new query is submitted, reset the pagination component to show the first page of results.

1. Submit a query which returns multiple pages of results
1. Click through to a result page other than 1
1. Submit a new query with multiple pages of results
1. Pagination component should show it's on the 1st page


packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -109,13 +118,13 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
...
             className={classNames.join(' ')}
-            page={page}
+            page={currentPage}
             totalItems={matchingResults}
             pageSize={actualPageSize}
             pageSizes={pageSizes}
             onChange={handleOnChange}
             itemRangeText={handleItemRangeText}
-            itemsPerPageText={mergedMessages.itemsPerPageText}
+            itemsPerPageText={'YAY'}
```
kabeaty: I think this should still be `mergedMessages.itemsPerPageText` and is also why some tests are failing...:)

examples/discovery-search-app/cypress/integration/pagination/spec.ts
```diff
@@ -104,6 +104,18 @@ describe('Pagination', () => {
           .find('.bx--select-input')
           .should('have.value', '2');
       });
+
+      describe('and we submit a new query', () => {
+        it.only('should return to the first page', () => {
```
kabeaty: I think you meant to remove the `only` here

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
ChrisGillin: doh :(

PR #94, opened 2020-02-12
fix: import individual react-virtualized components

jhpedemonte:

`rollup` is not able to tree-shake `react-virtualized` module, resulting in our build containing several RV components that we do not use.

Fix is to import only the components we need. This results in a reduction of ~130K in built size (according to `yarn analyze`). Also, it removes the `Import of non-existent export` warning.

Run build and `yarn analyze` before and after changes. Notice reduction in size and removal of warning.


jhpedemonte:
Since I was here, removed other RV warning. Fixes https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1110.

packages/discovery-react-components/package.json
```diff
@@ -58,6 +58,9 @@
     "react-virtualized": "^9.21.1"
   },
   "jest": {
+    "transformIgnorePatterns": [
+      "/!node_modules\\/react-virtualized/"
```
maniax89: do we want to include all node modules? or just this library?

yarn.lock
```diff
@@ -2256,7 +2256,7 @@
     "@hapi/hoek" "8.x.x"
 
 "@ibm-watson/discovery-react-components@file:packages/discovery-react-components":
-  version "1.1.0-beta.28"
+  version "1.1.0-beta.33"
```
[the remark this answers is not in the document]
jhpedemonte: Oh, hadn't noticed that. Not sure why it changed. I'll remove.

packages/discovery-react-components/package.json
jhpedemonte: This tells jest to not try to do a transform on this library, since it is already being included as an ES module. If it were commonjs, then it _would_ need to be transformed.

PR #97, opened 2020-02-17
chore: add test ids for use in discovery-tooling integration tests

ColbyJohnIBM:
- Add `data-testid` attributes to items that are used in document details integration tests

### DIFF
--- a/packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
+++ b/packages/discovery-react-components/src/components/CIDocument/components/CIDocument/CIDocument.tsx
@@ -406,7 +406,9 @@ const CIDocument: FC<CIDocumentProps> = ({
   return (
     <div className={base}>
       <nav className={`${base}__toolbar`} aria-label={messages.navigationToolbarLabel}>
-        <div className={`${base}__title`}>{filename}</div>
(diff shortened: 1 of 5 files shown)


H-Fin:
Is this the format we are using for test ids now?

noah-eigenfeld:
@H-Fin I think discovery components has always used the `data-testid` format for test ids. `data-test-id` was used in tooling for a little while, but after some discussion, we decided to unify to `data-testid`.

H-Fin:
@noah-eigenfeld I was asking about the string. I have previously been using the format `data-testid="filters-tab"`.

noah-eigenfeld:
> @noah-eigenfeld I was asking about the string. I have previously been using the format `data-testid="filters-tab"`.

My bad, haha!

Just did a quick search of the discovery components repo, and it looks like we don't have a set pattern for test id strings. I found some camelCased, some snake_cased, and kebab-cased was maybe the most prevalent. Might be a good idea to open up for discussion and see if there are strong feelings one way or another.

H-Fin:
@noah-eigenfeld Hmm, seems snake case was decided in a Slack thread [here.](https://ibm-analytics.slack.com/archives/CM2M88B6U/p1580934538494400)

PR #101, opened 2020-02-21
fix: standardize eslintrc

jhpedemonte:

Cleanup up `.eslintrc` and standardize with main repo. Fix linter warnings and errors.

Run lint command. If you want to get crazy, replace the use of `lodash.get` with optional chaining -- notice that linter doesn't complain about that any more.


jhpedemonte:
**Finally** was able to get optional chaining code passing the lint step. Required updating a bunch of dependencies, in particular the `prettier` ones.

Now I have issue with example app failing since that's built on CRA and now there are multiple different versions of some deps:

```
There might be a problem with the project dependency tree.

It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "babel-eslint": "10.0.3"

Don't try to install it manually: your package manager does it automatically.

However, a different version of babel-eslint was detected higher up in the tree:

  /home/travis/build/watson-developer-cloud/discovery-components/node_modules/babel-eslint (version: 10.1.0) 

Manually installing incompatible versions is known to cause hard-to-debug issues.

If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.

That will permanently disable this message but you might encounter other issues.

To fix the dependency tree, try following the steps below in the exact order:

  1. Delete package-lock.json (not package.json!) and/or yarn.lock in your project folder.

  2. Delete node_modules in your project folder.

  3. Remove "babel-eslint" from dependencies and/or devDependencies in the package.json file in your project folder.

  4. Run npm install or yarn, depending on the package manager you use.

In most cases, this should be enough to fix the problem.

If this has not helped, there are a few other things you can try:

  5. If you used npm, install yarn (http://yarnpkg.com/) and repeat the above steps with it instead.

     This may help because npm has known issues with package hoisting which may get resolved in future versions.

  6. Check if /home/travis/build/watson-developer-cloud/discovery-components/node_modules/babel-eslint is outside your project directory.

     For example, you might have accidentally installed something in your home folder.

  7. Try running npm ls babel-eslint in your project folder.

     This will tell you which other package (apart from the expected react-scripts) installed babel-eslint.

If nothing else helps, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.

That would permanently disable this preflight check in case you want to proceed anyway.
```

I think this is why @maniax89 was referring to the `react-app` lint config from deep within `node_modules`. But I'm trying to avoid that now. Some options for fixing this build error:
1. Ensure we have same versions of deps in `/package.json` as are required by the `react-scripts` dep in the example app.
    - While a simple enough fix, can lead to maintenance issues going forward as we'll need to keep these in sync.
2. Set `SKIP_PREFLIGHT_CHECK=true` during example build step to ignore this error.
    - As the warning message says, this _could_ lead to other (potentially hard to diagnose) errors.
3. ???

package.json
```diff
@@ -35,19 +42,27 @@
...
     "cross-env": "^5.1.4",
     "cypress": "3.8.2",
-    "eslint": "^6.8.0",
-    "eslint-config-prettier": "^6.2.0",
-    "eslint-plugin-prettier": "^3.1.0",
+    "eslint": "6.x",
+    "eslint-config-prettier": "^6.10.0",
+    "eslint-config-react-app": "^5.2.0",
+    "eslint-plugin-cypress": "^2.10.3",
+    "eslint-plugin-flowtype": "4.x",
```
maniax89: why are we using a mix of `2.x` and `^2.10.3`? can't we just pick one format?

.eslintrc
```diff
@@ -1,6 +1,14 @@
 {
-  "extends": ["./node_modules/react-scripts/node_modules/eslint-config-react-app/index.js"],
```
maniax89: yeah i don't know the best way to fix this. either `SKIP_PREFLIGHT_CHECK` or this i think are the only two options. I wasn't able to get resolutions to work properly

package.json
jhpedemonte: This was based on the command-line here: https://github.com/facebook/create-react-app/tree/master/packages/eslint-config-react-app#usage-outside-of-create-react-app. I'll change those versions to be consistent with rest of file.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -42,7 +41,7 @@ export const SimpleDocument: FC<Props> = ({
   if (document) {
     //Json object usually don't have enough info to allow us to determine which field to display
     //Unless there is passage pointing to a specific field.
-    const isJsonType = get(document, 'extracted_metadata.file_type') === 'json';
+    const isJsonType = document?.extracted_metadata?.file_type === 'json';
```
jhpedemonte: This now works with this PR. Previously, would fail in lint step.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -51,9 +50,14 @@ export const SimpleDocument: FC<Props> = ({
       if (highlight && isPassage(highlight)) {
         passage = highlight as QueryResultPassage;
         field = passage.field;
+        if (!field) {
+          // if passage has no defined field, choose a default and unset `highlight`
+          field = 'text';
+          highlight = undefined;
+        }
```
jhpedemonte: Since the SDK defines `field` as an optional prop for `QueryResultPassage`, had to add this fallback code.

.eslintrc
jhpedemonte: Decided to go with `SKIP_PREFLIGHT_CHECK`. Seems to be working, for now.

packages/discovery-react-components/package.json
```diff
@@ -71,5 +71,9 @@
       "!src/**/__stories__/**",
       "!src/**/*.stories.*"
     ]
+  },
+  "devDependencies": {
```
maniax89: do we use these in discovery-react-components?

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
maniax89: can we get rid of lodash/get altogether?

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
maniax89: how was this working before?

packages/discovery-react-components/package.json
jhpedemonte: Unsure why, but I had to install these in order to get Storybook to run/build. Unsure why they weren't needed before.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
jhpedemonte: Yes, we can. But best to handle as separate PR -- this was just meant to be a demonstration of something that previously failed but now works with this PR.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
jhpedemonte: Don't know. I was wondering the same thing. Might be as a result of updated dependencies, which are now working better at detecting certain issues.

packages/discovery-react-components/src/components/CIDocument/CIDocument.tsx
```diff
@@ -1,20 +1,20 @@
 import CIDocument, {
-  CIDocumentProps,
+  CIDocumentProps as _CIDocumentProps,
```
broulaye: why is this change necessary

jhpedemonte: Because this file contains both an instance and a type named `CIDocumentProps`. It looks like this was "allowed" by previous versions of TypeScript, but now it complains.

ColbyJohnIBM:
We are waiting to merge this until after the March Fixpack.

I will note this in the title of this PR as well.

PR #102, opened 2020-02-27
fix: inconsistent document naming

broulaye:

Fixing the inconsistent document naming between SearchResults and DocumentDetails, by first trying to display the document title if it's available, otherwise display the file name, otherwise default to the document id. Since this was already being done on the component side of things, I decided to make an utility function out of it and export it for the tooling to use.


packages/discovery-react-components/src/components/SearchResults/utils/__tests__/getDocumentTitle.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+const mockResultWithFilename = {
+  document_id: 'df4c085040a9049ddf956ca6a213699d',
+  extracted_metadata: {
+    file_type: 'html',
+    filename: 'Family Day 2020 (Israel)'
+  }
+};
+
+describe('getDocumentTitle', () => {
+  test('returns result title', () => {
```
H-Fin: Is there a reason you are using `test` instead of `it`?

packages/discovery-react-components/src/components/SearchResults/utils/__tests__/getDocumentTitle.test.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { getDocumentTitle } from '../index';
+
+const mockResultWithTitle = {
```
H-Fin: In order to increase readability, we could move these constants to sit right above their tests.

packages/discovery-react-components/src/components/SearchResults/utils/__tests__/getDocumentTitle.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    );
+  });
+
+  test('returns result document id', () => {
+    expect(getDocumentTitle('title', mockResultWithoutTitle)).toEqual(
+      mockResultWithoutTitle.document_id
+    );
+  });
+
+  test('returns result file name', () => {
```
H-Fin: Might be helpful to move this test above the document id one since the order is `title -> filename -> document_id`. You could also note in the test string something like: `it('returns result filename if title is unavailable')`

packages/discovery-react-components/src/components/SearchResults/utils/__tests__/getDocumentTitle.test.tsx
broulaye: to be consistent with other test in component

packages/discovery-react-components/src/components/SearchResults/utils/__tests__/getDocumentTitle.test.tsx
broulaye: sure thing 👍

broulaye:
@JuanNicolasGarcia Those have already been added. I forgot to add any criteria but you could link it and test it on your end to make sure it works.

packages/discovery-react-components/src/components/SearchResults/utils/getDocumentTitle.ts
```diff
@@ -0,0 +1,18 @@
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
+import get from 'lodash/get';
+
+export const getDocumentTitle = (
+  resultTitleField: string,
+  result?: DiscoveryV2.QueryResult
+): string => {
```
jhpedemonte: Switch order of params: the main object first, followed by the "qualifier".

Also, just because they were called "result" and "resultTitleField" in the earlier code doesn't mean we have to use those same names here. This is a standalone function and "result" makes no sense here. Use something like `document` and `titleField` instead.

packages/discovery-react-components/src/components/SearchResults/utils/getDocumentTitle.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (result) {
+    const documentId: string | undefined = get(result, 'document_id');
+    const filename: string | undefined = get(result, 'extracted_metadata.filename');
+    const userSelectedTitle: string | undefined = get(result, resultTitleField);
+    const title: string = userSelectedTitle
+      ? userSelectedTitle
+      : get(result, 'extracted_metadata.title') || filename || documentId;
+    return title;
+  }
+  return 'No Title';
```
jhpedemonte: I don't think we'll ever hit this (why show a title if we have no document?), but I don't like this default. Either return `null` or an empty string (I prefer the former).

packages/discovery-react-components/src/components/SearchResults/utils/getDocumentTitle.ts
```diff
@@ -0,0 +1,18 @@
...
+  result?: DiscoveryV2.QueryResult
+): string => {
+  if (result) {
+    const documentId: string | undefined = get(result, 'document_id');
+    const filename: string | undefined = get(result, 'extracted_metadata.filename');
+    const userSelectedTitle: string | undefined = get(result, resultTitleField);
+    const title: string = userSelectedTitle
+      ? userSelectedTitle
+      : get(result, 'extracted_metadata.title') || filename || documentId;
+    return title;
```
jhpedemonte: There's a lot of unnecessary code, some duplication and types that are getting in the way and making this much more difficult to read than necessary.

```suggestion
    return result[resultTitleField] || 
        get(result, 'extracted_metadata.title') ||
        get(result, 'extracted_metadata.filename') ||
        result.document_id;
```

packages/discovery-react-components/src/components/SearchResults/utils/getDocumentTitle.ts
jhpedemonte: I think `titleField` should be optional in this function. So default (if no "overriding" field is specified) is: extracted title, file name, doc ID.

packages/discovery-react-components/src/index.tsx
```diff
@@ -14,3 +14,4 @@ export {
 } from './components/DocumentPreview/components/HtmlView/HtmlView';
 export { CIDocument, canRenderCIDocument } from './components/CIDocument/CIDocument';
 export { default as StructuredQuery } from './components/StructuredQuery/StructuredQuery';
+export { getDocumentTitle } from './components/SearchResults/utils';
```
jhpedemonte: We need to think about how we're exporting things from components, since these are our API. We have the top level components, but utility methods like this one should probably be namespaced somehow. Let's discuss with rest of team tomorrow.

packages/discovery-react-components/src/components/SearchResults/utils/getDocumentTitle.ts
broulaye: for the first comment, the reason I had it in this order is because typescript doesn't like it when you have required parameter follow an optional parameter.

jhpedemonte: Then either change it to be not optional (since that will be quite rare anyway, if at all) or change it's type to `QueryResult | null`.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -20,6 +20,7 @@ import {
   searchResultFooterTitleClass,
   searchResultFooterCollectionNameClass
 } from 'components/SearchResults/cssClasses';
+import { getDocumentTitle } from '../../utils';
```
kabeaty: In this case, I'd go with importing with the module path alias `components/SearchResults/utils`.

packages/discovery-react-components/src/components/SearchResults/utils/getDocumentTitle.ts
```diff
@@ -0,0 +1,18 @@
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
```
jhpedemonte: I think we should have this file at the top-level `utils` folder, not within `SearchResults/`

broulaye: wouldn't it best to leave it here since it only concern the `SearchResults` and unlikely to be used anywhere else

jhpedemonte: Well, in this case we're looking to export it as part of our public API, and not as a member of `SearchResults`. For that reason, I think our code layout should match our API.

broulaye: Fair enough 👍

broulaye:
@jhpedemonte this is ready for a re-review

packages/discovery-react-components/src/utils/getDocumentTitle.ts
```diff
@@ -0,0 +1,17 @@
+import DiscoveryV2 from 'ibm-watson/discovery/v2';
+import get from 'lodash/get';
+
+export const getDocumentTitle = (
+  result: DiscoveryV2.QueryResult | undefined,
+  resultTitleField: string
```
jhpedemonte: You didn't rename these arguments, as I commented on previously.

PR #103, opened 2020-03-02
fix: search facet text not wrapping

H-Fin:
Allows the search facet text to wrap onto the next line if the string is too long. Some styling changes are also added as a PR in the tooling repo for the radio buttons.
**Before:**
Screen Shot 2020-02-28 at 2 50 28 PM
**After:**
Screen Shot 2020-03-02 at 12 10 24 PM

- Link this PR with the corresponding one in tooling.
- In the Improve and Custom section of a project add Customize Display facets with longer strings.
- See that the overflow text wraps.

No

Connecting issue: https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1201


packages/discovery-styles/scss/components/search-facets/_search-facets.scss
```diff
@@ -10,6 +10,7 @@
 
 .#{$prefix}--search-facet {
   margin-bottom: $spacing-md;
+  word-break: break-word;
```
jhpedemonte: If I'm reading the screenshots correctly, this will also break a very long word, correct?

PR #104, opened 2020-03-04
feat: add DocumentPreview error message

H-Fin:
Adds an error message to document preview for when JSON and CSV files do not have a passage to load. 

https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1192

**Before for CSV:**
Screen Shot 2020-03-04 at 4 46 29 PM

**After for CSV and JSON:**
Screen Shot 2020-03-05 at 12 39 11 PM

On Example App:
1. Run the Components Example App against a project with a CSV file and do NOT specify a search result field.
2. See that the error message appears in the Document tab.
3. On the cluster, set a search result field. 
4. On the Example App, see that the error message is gone and the field is shown.

On Cluster:
1. Link components and tooling. Make sure you also link components styles.
2. Look at a project with a CSV file and NOT a specified search result field.
3. See error message on Document tab.
4. Specify a search result field, go back to Document tab and see that the error message is gone.

Tests have been updated to reflect this change.

No


examples/discovery-search-app/src/app.scss
```diff
@@ -116,19 +116,10 @@ $nav-toolbar-height: 42px;
     top: 0;
   }
 
-  .#{$prefix}--document-preview {
```
H-Fin: I removed some stylings in the Discovery Search App so that it resembles Tooling more. It had a lot of strange margins and padding.

packages/discovery-react-components/src/components/DocumentPreview/__tests__/DocumentPreview.test.tsx
```diff
@@ -305,13 +305,12 @@ describe('DocumentPreview', () => {
   };
 
   describe('with CSV files', () => {
-    // TODO: Currently fails since no error message is displayed if there is an unspecified body field and passage
-    it.skip('should show an error if no "body" field or passage have been specified', () => {
+    it('should show an error if no "body" field or passage have been specified', () => {
```
jhpedemonte: 🎉

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/ErrorView.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <div className={base} data-testid="error_state">
+      <div className={`${base}__icon`}>
+        <DocumentPreviewIcon />
+      </div>
+      <div>
+        <h1 className={`${base}__header`}>Can’t preview document</h1>
+        <div className={`${base}__message`}>
+          Try the JSON tab for a different view of this document's data.
+        </div>
```
jhpedemonte: The way we handle i18n messages is to allow them to be overridden by the user. They are defined as optional params, with default values.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -31,8 +32,7 @@ export const SimpleDocument: FC<Props> = ({
   document,
   highlight,
   setLoading,
-  setHideToolbarControls,
-  cannotPreviewMessage = 'Cannot preview document'
```
jhpedemonte: Like here: `cannotPreviewMessage` is the overrideable (sp?) "i18n" message. You'll need to add another one for the sub-heading.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -41,12 +41,18 @@ export const SimpleDocument: FC<Props> = ({
   let html,
     passage: QueryResultPassage | null = null;
   if (document) {
-    // JSON files will default to displaying the specified body field, `text` field, or passage highlighting field,
+    // JSON and CSV files will default to displaying the specified body field, `text` field, or passage highlighting field.
     // Otherwise an error is shown
     const isJsonType = isJsonFile(document);
+    const isCsvType = isCsvFile(document);
```
jhpedemonte: To make this a bit cleaner, just put both into a single var:

```suggestion
    const isCsvOrJson = isCsvFile(document) || isJsonFile(document);
```

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -139,7 +145,11 @@ export const SimpleDocument: FC<Props> = ({
         />
       </div>
     </div>
-  ) : null;
+  ) : (
+    <div className={base}>
+      <ErrorView />
+    </div>
+  );
```
jhpedemonte: Don't duplicate the wrapper div:

```suggestion
  return (
      <div className={base}>
          {html ? {
              <div className={`${base}__wrapper`}> ...
          ) : (
              <ErrorView />
          )}
      </div>
  );
```

packages/discovery-styles/scss/components/document-preview/_simple-document.scss
```diff
@@ -20,3 +20,33 @@
 .#{$prefix}--simple-document__content {
   max-width: 78ch;
 }
+
+$width: 500px;
```
jhpedemonte: Why? Where did this number come from?

packages/discovery-styles/scss/components/document-preview/_simple-document.scss
```diff
@@ -20,3 +20,33 @@
...
+  display: inline-block;
+  width: $icon-size;
+  height: $icon-size;
+  margin-right: $spacing-lg;
+}
+
+.#{$prefix}--simple-document__error-view__header {
+  font-size: 1.25rem;
+  padding-bottom: $spacing-sm;
+  color: #171717; // primary text color in tooling
```
jhpedemonte: We should be able to use a var here and below.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/ErrorView.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const ErrorState = () => {
+  const base = `${settings.prefix}--simple-document__error-view`;
+  return (
+    <div className={base} data-testid="error_state">
+      <div className={`${base}__icon`}>
+        <DocumentPreviewIcon />
+      </div>
+      <div>
+        <h1 className={`${base}__header`}>Can’t preview document</h1>
```
anibapat: Could you add i18n here and on line 15?

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/ErrorView.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { settings } from 'carbon-components';
+import { ReactComponent as DocumentPreviewIcon } from './icons/document_preview.svg';
+
+const ErrorState = () => {
+  const base = `${settings.prefix}--simple-document__error-view`;
```
kabeaty: What about calling this `baseClass` instead of simply `base`? I think that this is clearer and also aligns with what we are using in other components.

packages/discovery-styles/scss/components/document-preview/_simple-document.scss
```diff
@@ -20,3 +20,33 @@
 .#{$prefix}--simple-document__content {
   max-width: 78ch;
 }
+
+$width: 500px;
+$icon-size: 80px;
```
kabeaty: I think it would be helpful for readability to define these at the top of the file, and this is also consistent with `_search-results.scss`, for example

packages/discovery-styles/scss/components/document-preview/_simple-document.scss
```diff
@@ -20,3 +20,33 @@
...
+
+.#{$prefix}--simple-document__error-view {
+  width: $width;
+  padding-top: 15vh;
```
kabeaty: Could just combine these into one `padding: 15vh 0`

H-Fin:
> Looks good overall! Left a few comments. I'm also seeing in the example app that the header doesn't look as bold as the screenshot you included, and the spacing looks weird to me around the `x` in the upper right (not sure if this is related to a change in this PR). I didn't test in Tooling, just took a look in the example app. Screenshot here:
> Screen Shot 2020-03-05 at 10 48 26 AM

My screenshot is from Tooling which is why things are a bit different. I will investigate the 'x' and bolding on the example app to see if my changes affected it. Thanks for the review!

kabeaty:
Sounds good, thanks! To clarify since I think it might not have been clear, by the heading, I meant `Can't preview document` in the error message. Yeah, I just figured we'd probably want to style similar to Tooling in these instances, even though there will be some differences between them.

packages/discovery-styles/scss/components/document-preview/_simple-document.scss
H-Fin: It is the same width the EmptyState component uses in tooling. The message width ends up being shorter than 500px though, so its not doing anything here--but if the message changes to something longer in the future it might be helpful. I'm hesitant to add a comment stating its the width from Tooling since this is an open source component. Thoughts?

anibapat:
@H-Fin 
>My screenshot is from Tooling which is why things are a bit different. I will investigate the 'x' and bolding on the example app to see if my changes affected it. Thanks for the review!

I think you need to pull down the latest for `tooling`. We have changed the style of Header, you won't see the `X` button there, we have replaced that with a back button. as part of this: https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1078

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -32,7 +34,8 @@ export const SimpleDocument: FC<Props> = ({
   highlight,
   setLoading,
   setHideToolbarControls,
-  cannotPreviewMessage = 'Cannot preview document'
+  cannotPreviewHeader = `Can’t preview document`,
+  cannotPreviewMessage = `Try the JSON tab for a different view of this document's data.`
```
anibapat: There should be single quotes `' '` (not the backquotes)

packages/discovery-styles/scss/components/document-preview/_simple-document.scss
jhpedemonte: I suggest creating a var like `$empty_state_message_width` and adding to `_vars.scss`. That way, it's top-level and not buried inside one of the components.

jhpedemonte:
1. The close icon is not related to this PR. I've seen that before. Should be handled elsewhere.
2. @anibapat The screenshots above are for the example app, which is still using a simple modal. So not relevant to tooling changes.

anibapat:
Sorry, I think I misunderstood with the header you have in the `example app` vs what we have in the Tooling.

packages/discovery-react-components/src/components/DocumentPreview/__tests__/DocumentPreview.test.tsx
```diff
@@ -246,7 +246,7 @@ describe('DocumentPreview', () => {
         ({ getByText } = render(<DocumentPreview document={errorJsonDoc} />));
       });
 
-      getByText('Cannot preview document');
+      getByText("Can't preview document");
```
anibapat: I wonder how prettier did not catch the double-quotes. Here and on line 313

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -32,7 +34,8 @@ export const SimpleDocument: FC<Props> = ({
   highlight,
   setLoading,
   setHideToolbarControls,
-  cannotPreviewMessage = 'Cannot preview document'
+  cannotPreviewHeader = "Can't preview document",
+  cannotPreviewMessage = "Try the JSON tab for a different view of this document's data."
```
anibapat: Double quotes.

packages/discovery-react-components/src/components/DocumentPreview/__tests__/DocumentPreview.test.tsx
H-Fin: I used single quotes here but my linter has changed it. Just tried again and same thing.

H-Fin:
@kabeaty I went ahead and changed the `font-weight` of the example app `h1` to match that of Tooling. It should be more similar now.
Screen Shot 2020-03-05 at 1 09 49 PM

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/ErrorView.tsx
kabeaty: I see this updated in `SimpleDocument` but not here? And thanks for updating in the other file btw.

H-Fin: Good catch--fixed!

packages/discovery-react-components/src/components/DocumentPreview/__tests__/DocumentPreview.test.tsx
H-Fin: Our ESLint rules currently change it to double quotes if there is an escaped character inside the string.

packages/discovery-react-components/src/components/DocumentPreview/components/SimpleDocument/SimpleDocument.tsx
```diff
@@ -32,7 +34,8 @@ export const SimpleDocument: FC<Props> = ({
   highlight,
   setLoading,
   setHideToolbarControls,
-  cannotPreviewMessage = 'Cannot preview document'
+  cannotPreviewHeader = 'Can\'t preview document',
+  cannotPreviewMessage = 'Try the JSON tab for a different view of this document\'s data.'
```
jhpedemonte: This is part of our API, so what you are proposing is a breaking change.

I suggest keeping the "header" as `cannotPreviewMessage` and name the other prop as something like `cannotPreviewMessage2`

PR #108, opened 2020-03-11
fix: callbacks optimizations

jhpedemonte:

Remove need to disable linting check for `useEffect` dependencies when it comes to callbacks.

### DIFF
--- a/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
+++ b/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, FC, useEffect, useState, useMemo, useCallback } from 'react';
+import React, { createContext, FC, useEffect, useState, useCallback } from 'react';
 import DiscoveryV2 from 'ibm-watson/discovery/v2';
 import {
   useDeepCompareEffect,
@@ -354,7 +354,7 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
         setAutocompletions(null);
       }
     },
-    [autocompletionOptions, projectId, searchClient]
+    [autocompletionOptions, projectId]
   );
 
   const handleFetchAggregations = useCallback(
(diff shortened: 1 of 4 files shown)


packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -196,12 +198,13 @@ export const useFieldsApi = (
     searchClient.listFields,
     searchClient
   );
+
+  const fetchFields = useCallback(() => setFetchToken({ trigger: true }), [setFetchToken]);
```
jhpedemonte: These type of callbacks were getting recreated on every render (since they were anonymous functions created within an object definition). Instead, wrap them with `useCallback` to prevent re-creation.

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -417,26 +417,17 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
+    fetchAutocompletions: handleFetchAutocompletions,
+    fetchDocuments: handleFetchDocuments,
+    setSelectedResult: handleSetSelectedResult,
+    setAutocompletionOptions,
     setSearchParameters,
-    handleFetchFields
-  ]);
+    setIsResultsPaginationComponentHidden,
+    fetchFields: handleFetchFields
+  };
```
jhpedemonte: Most of these functions are either already a result of `useCallback` or `useState`, so no need to wrap with `useMemo`.

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -96,7 +96,7 @@ const SearchFacets: FC<SearchFacetsProps> = ({
   useEffect(() => {
     fetchAggregations(searchParameters);
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
maniax89: should we remove the eslint disable here?

broulaye: nope the one here is necessary, otherwise `eslint` will automatically add `searchParameters` to the dependencies which breaks some functionalities

maniax89: I thought that the point of this PR was to remove the linter ignores?

broulaye: Yes, but in some cases like this one it’s actually necessary I believe. Since we don’t want to include the whole ‘searchParameters’ as dependency just one of its values

PR #110, opened 2020-03-19
docs: update release docs w/ instructions for storybook build

jhpedemonte:

Adds step to build storybook in release steps.

Implements https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1166

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,7 @@ We are adopting a ["oneflow" branching strategy](https://www.endoflineblog.com/o
   - `git merge --ff-only @ibm-watson/discovery-react-components@1.0.1`
 - `release/x.x.x` is a temporary branch created for beginning a production release. this contains all the features needed for the release and will only receive bugfixes. Once the release is complete, this branch is tagged and merged back into `develop`. example steps:
   - `git checkout release/2.3.0`
+  - `yarn install && yarn workspace @ibm-watson/discovery-components run storybook:build`
   - if we want to publish a release candidate (not final build):
     - `npx lerna publish --conventional-prerelease --preid rc --dist-tag rc`
     - (after we find out the `rc` is good to go) `npx lerna publish --create-release github --conventional-graduate` [docs](https://github.com/lerna/lerna/blob/master/commands/version/README.md#--conventional-graduate)


README.md
```diff
@@ -430,6 +430,7 @@ We are adopting a ["oneflow" branching strategy](https://www.endoflineblog.com/o
   - `git merge --ff-only @ibm-watson/discovery-react-components@1.0.1`
 - `release/x.x.x` is a temporary branch created for beginning a production release. this contains all the features needed for the release and will only receive bugfixes. Once the release is complete, this branch is tagged and merged back into `develop`. example steps:
   - `git checkout release/2.3.0`
+  - `yarn install && yarn workspace @ibm-watson/discovery-components run storybook:build`
```
maniax89: if we are putting this step in here, don't we want to remove it from the `prepublish` step? -> https://github.com/watson-developer-cloud/discovery-components/blob/develop/packages/discovery-react-components/package.json#L25

Also i would like to adopt potentially having a `gh-pages` branch instead of putting everything in `/docs`

jhpedemonte: @maniax89 `prepublish` does runs the `build` script, which only does rollup (I assume that it _does_ do `yarn install`, or that happens before). Should we add `yarn run storybook:build` as part of `prepublish` also?

maniax89: sorry i guess i meant to say that `yarn storybook:build` is already part of `preversion`, so it gets run alongside the versioning step to make sure that the version matches the storybook diff.

we can avoid this completely if we publish the storybook versioning to a separate `gh-pages` branch which would be more resilient anyways.

`yarn install` is _not_ being run, nor is any sort of `clean` which we should probably add as well to the `prepublish` script

jhpedemonte: @maniax89 So for the latter issue, does this seem correct:

```
diff --git a/packages/discovery-react-components/package.json b/packages/discovery-react-components/package.json
index 9f10de9..4ab1e8a 100644
--- a/packages/discovery-react-components/package.json
+++ b/packages/discovery-react-components/package.json
@@ -14,6 +14,7 @@
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
     "circular": "madge --circular src/*",
     "build": "rollup -c",
+    "clean": "rimraf dist/",
     "eslint": "eslint --resolve-plugins-relative-to '../../node_modules/react-scripts/node_modules' './{src,.storybook}/**/*.{js,jsx,ts,tsx}' --quiet",
     "lint": "yarn run eslint && yarn run circular",
     "start": "rollup -c -w",
@@ -21,7 +22,7 @@
     "storybook:build": "build-storybook",
     "storybook:build:release": "STORYBOOK_BUILD_MODE=production build-storybook -o ../../docs/storybook",
     "analyze": "source-map-explorer 'dist/index.js'",
-    "prepublish": "yarn run build",
+    "prepublish": "yarn run clean && yarn run build",
     "preversion": "yarn run storybook:build:release",
     "version": "git add ../../docs"
   },
```

And as you said, no need to add something to README for building storybook.

---



> we can avoid this completely if we publish the storybook versioning to a separate `gh-pages` branch which would be more resilient anyways.

What do you have in mind for this?

maniax89: what i was thinking for the `gh-pages` branch to run as a separate thing as part of the "deploy" step instead of munging it in with the `preversion` step. that way our `gh-pages` branch can be dedicated to the storybook dist code since its sort of meaningless for diffs in the `docs` folder

maniax89:
closing due to inactivity

PR #114, opened 2020-03-26
fix: add result count to search facets

H-Fin:
Adds the number of found documents to each search facet label. The search facets are ordered in descending order. Does not show a count for Dynamic Facets as those are not supported by the API yet.
ezgif com-video-to-gif (2)

https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/911

- Run `yarn storybook` inside `discovery-react-components`, click on Search Facets and see the document counts are now there.
- Or link components to `discovery-tooling` and see the search facets now have result counts.

Cypress tests in the example app have been updated.

No

### DIFF
--- a/examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
+++ b/examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
@@ -29,7 +29,7 @@ describe('Multi-Select Facets', () => {
         .as('priceFacet');
     });
 
-    it('available facets are listed on the side', () => {
+    it('shows available facets listed on the side', () => {
       cy.get('.bx--search-facet')
         .as('allFacets')
         .should('have.length', 4);
@@ -38,7 +38,7 @@ describe('Multi-Select Facets', () => {
       cy.get('@priceFacet').should('exist');
     });
 
-    it('facets are displayed as checkboxes', () => {
+    it('has facets displayed as checkboxes', () => {
       cy.get('@cityFacet')
         .find('.bx--checkbox')
         .should('have.length', 5);
@@ -50,7 +50,7 @@ describe('Multi-Select Facets', () => {
         .should('have.length', 3);
     });
 
-    it('facets have matching results count', () => {
+    it('shows the facet matching results count', () => {
       cy.get('@cityFacet')
         .contains('Boston, MA (30000)')
         .should('exist');
@@ -74,7 +74,7 @@ describe('Multi-Select Facets', () => {
           .should('eq', 'location:"Ames, IA"');
       });
 
-      it('the bubble next to that facet says 1', () => {
+      it('changes the bubble next to that facet to say 1', () => {
         cy.get('.bx--list-box__selection')
           .contains('1')
           .should('exist');
@@ -98,7 +98,7 @@ describe('Multi-Select Facets', () => {
             .should('eq', 'location:"Pittsburgh, PA"|"Ames, IA"');
         });
 
-        it('the bubble next to that facet says 2', () => {
+        it('changes the bubble next to that facet to say 2', () => {
           cy.get('.bx--list-box__selection')
             .contains('2')
             .should('exist');
@@ -121,7 +121,7 @@ describe('Multi-Select Facets', () => {
             .should('eq', '');
         });
 
-        it('the "Clear all" button disappears', () => {
+        it('has the "Clear all" button disappear', () => {
           cy.get('button')
             .contains('Clear all')
             .should('not.exist');
@@ -174,7 +174,7 @@ describe('Multi-Select Facets', () => {
             cy.wait('@postQueryFacetsAmes').as('amesFilterQueryObject');
           });
 
-          it('the filters from only that facet are cleared', () => {
+          it('has the filters from only that facet clear', () => {
             cy.get('@amesFilterQueryObject')
               //@ts-ignore TODO: we'll need to handle typings for `cy.its` at some point, but for now, we'll ignore the error on the parameter string
               .its('requestBody.filter')
--- a/examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
+++ b/examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
@@ -50,6 +50,12 @@ describe('Multi-Select Facets', () => {
         .should('have.length', 3);
     });
 
+    it('facets have matching results count', () => {
+      cy.get('@cityFacet')
+        .contains('Boston, MA (30000)')
+        .should('exist');
+    });
+
     describe('and a facet filter is selected', () => {
       beforeEach(() => {
         cy.route('POST', '**/query?version=2019-01-01', '@facetsQueryAmesJSON').as(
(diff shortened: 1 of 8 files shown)


packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/SingleSelectFacetsGroup.tsx
```diff
@@ -65,14 +65,20 @@ export const SingleSelectFacetsGroup: FC<SingleSelectFacetsGroupProps> = ({
     >
       {facets.map(facet => {
         const text = get(facet, facetsTextField, '');
+        const matchingResults = get(facet, 'matching_results');
+        let fullText = text;
```
anirudhaBapat: Why not use `text` (instead of a new variable)

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/SingleSelectFacetsGroup.tsx
```diff
@@ -65,14 +65,20 @@ export const SingleSelectFacetsGroup: FC<SingleSelectFacetsGroupProps> = ({
     >
       {facets.map(facet => {
         const text = get(facet, facetsTextField, '');
+        const matchingResults = get(facet, 'matching_results');
```
anirudhaBapat: want to use optional chaining as it is available now?

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
```diff
@@ -56,6 +56,12 @@ export const MultiSelectFacetsGroup: FC<MultiSelectFacetsGroupProps> = ({
     <>
       {facets.map(facet => {
         const text = get(facet, facetsTextField, '');
+        const matchingResults = get(facet, 'matching_results');
+        let fullText = text;
+        // To prevent showing the number of Dynamic Facets until the API supports this
+        if (matchingResults > 0) {
```
maniax89: One reason we didn't implement this initially is due to how we end up going through and selecting facets. I would try using it a bit with multiple facets to see what ends up showing for the matching results after selecting across facet groups and make sure that doesn't look strange (specifically, whether or not the numbers change or go to 0 when we would/wouldn't expect them to)

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/SingleSelectFacetsGroup.tsx
H-Fin: That value is used both for the `data-key` and buffer below which need a version of the label without the number.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
H-Fin: In my experience playing with the facets, the matching results value never changes. Each selected facet adds its matching results to the list of documents shown--it does not filter down.

examples/discovery-search-app/cypress/integration/multi_select_facets/spec.ts
```diff
@@ -50,6 +50,12 @@ describe('Multi-Select Facets', () => {
         .should('have.length', 3);
     });
 
+    it('facets have matching results count', () => {
```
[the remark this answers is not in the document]
broulaye: I see other tests here are doing the same thing you're doing, so you don't have to change it, but if you do I would say maybe change the other ones too to follow that rule.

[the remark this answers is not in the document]
H-Fin: Agreed! Went ahead and updated the test descriptions in all three files to read more as sentences.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
maniax89: @H-Fin that's something i would bring up with design -> do we want those numbers to change?

the implementation we went with was one that allowed you to expand selections after 2 or more facet groups are selected.

the alternative is to filter down the list of available facets to the ones returning on the request to indicate how your filter affected the result set. the downside of this approach is that exploratory faceting is mostly eliminated without undoing selections (no way to say "i want "x" OR "y" instead of just "x" AND "y")

maniax89: also - i think we order by these `matching_results` but i faintly remember also bubbling the selected ones to the top, does that happen? does it look weird if the numbers are out of order?

H-Fin: We don't bubble selected ones to the top. As of right now, after each search query, the non-dynamic facets do not change order or appear/disappear based off facet filtering.

maniax89: got it - must be the dynamic facets that do that

H-Fin: I am investigating what the expected behavior is of the other facets though and will update accordingly.

H-Fin: A separate issue will address facet groups (#1352). I added a gif to this PR so people can see the current behavior more easily.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
```diff
@@ -56,6 +56,9 @@ export const MultiSelectFacetsGroup: FC<MultiSelectFacetsGroupProps> = ({
     <>
       {facets.map(facet => {
         const text = get(facet, facetsTextField, '');
+        const matchingResults = facet.matching_results;
+        const labelText =
+          matchingResults !== undefined ? text + ' (' + matchingResults + ')' : text;
```
jhpedemonte, line 61: This is not translatable. I think best option is to have two different i18n message strings:
- `{facetText}`
- `{facetText} ({count})`

Then choose correct one and substitute values.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/SingleSelectFacetsGroup.tsx
```diff
@@ -65,14 +65,17 @@ export const SingleSelectFacetsGroup: FC<SingleSelectFacetsGroupProps> = ({
     >
       {facets.map(facet => {
         const text = get(facet, facetsTextField, '');
+        const matchingResults = facet.matching_results;
+        const labelText =
+          matchingResults !== undefined ? text + ' (' + matchingResults + ')' : text;
```
jhpedemonte, line 70: Same as above.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
H-Fin: This should be resolved now

PR #115, opened 2020-03-27
feat: identify Top Entities facet values by Type

kabeaty:
These changes add in a call with a nested aggregation query to get back the type for entities facets so we can use it to identify/group these facets' facet values by type. 

Any entities facet will be displayed with its facet values organized by type (I'm using `category` in the code to distinguish from types), so types like `Company`, `Person`, `Organization`, etc. On initial load, only the types will be shown, collapsed, beneath the entities facet's label. By clicking on them, they will expand to show their facet values. If there are more than 5 facet values, a `Show more` button will appear that will reveal the rest of the facet values if clicked. The facet values can be selected and cleared like any other facet. 

See screenshot here:
Screen Shot 2020-04-03 at 5 18 22 PM

You can link to Tooling and test in a workspace there. Try out some searches on a workspace with an entities facet and see that the types show up as expected. Play around with expanding/collapsing categories and facet values within categories. You can also test out in the example app with some searches there on a project that has an entities facet (Sample Project has a `Top Entities` by default).

You can also check it out in Storybook.

Also check out the tests that I added, and confirm they're testing everything we'd want to.

An example is displayed in Storybook for `SearchFacets` component.

No


packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
```diff
@@ -32,3 +38,9 @@ export interface CollectionItem {
   id: string;
   label: string;
 }
+
+export interface QueryAggregationWithName extends DiscoveryV2.QueryAggregation {
+  field?: string;
+  count?: number;
+  name?: string;
```
maniax89: Doesn't he SDK already define `name`? If not, we should be pushing that into the SDK

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
```diff
@@ -1,12 +1,42 @@
 import DiscoveryV2 from 'ibm-watson/discovery/v2';
+import { QueryAggregationWithName } from './searchFacetInterfaces';
+
+const TOP_ENTITIES_FIELD = 'enriched_text.entities.text';
+const ENTITIES_NESTED_TYPE_TERM_AGG = '.term(enriched_text.entities.type,count:1)';
 
 export const buildAggregationQuery = (
-  configuration: DiscoveryV2.QueryTermAggregation[]
+  configuration?: QueryAggregationWithName[],
```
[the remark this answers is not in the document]
kabeaty: Refactored to just use the `configuration` for this and so now no longer optional 👍

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
```diff
@@ -1,12 +1,42 @@
...
+      const termCount = term.count ? ',count:' + term.count : '';
+      const termName = term.name ? ',name:' + term.name : '';
+      const topEntitiesNestedTypeTermAgg =
+        term.field === TOP_ENTITIES_FIELD ? ENTITIES_NESTED_TYPE_TERM_AGG : '';
+      return 'term(' + term.field + termCount + termName + ')' + topEntitiesNestedTypeTermAgg;
+    });
+    return '[' + aggregation!.toString() + ']';
+  } else {
+    if (searchParamsAgg.includes(TOP_ENTITIES_FIELD)) {
+      const searchParamsAggArr = searchParamsAgg
```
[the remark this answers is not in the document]
kabeaty: Yes, agree...the way I had originally gotten this working, it was needed, but now we can just use the structured response to handle this. Updated, thanks!

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
kabeaty: The SDK defines `name` in `ComponentSettingsAggregation` but not for the query aggregation. Yes, I wanted to get something working here at least, but I agree it should be defined in the SDK. Should I just open a PR here (https://github.com/watson-developer-cloud/node-sdk) to add it in?

maniax89: We should probably push it to the API definitions repo so it gets added to all languages

H-Fin:
Verified it works locally and LGTM. One small note:

When tabbing through, I noticed that there is nothing to indicate when the facet chevron is focused on but other chevrons on the page have a focus state. Do we want to add one?
Screen Shot 2020-04-13 at 1 44 33 PM

ColbyJohnIBM:
The filter panel counts do not always match the search results count when a filter is applied. We should file another issue in `wire-plugin-tracker`, like this one: https://github.ibm.com/Watson-Discovery/wire-plugin-tracker/issues/591

Screen Shot 2020-04-14 at 10 01 24 AM
Screen Shot 2020-04-14 at 10 01 38 AM

packages/discovery-react-components/src/components/SearchFacets/components/__tests__/FieldFacets.test.tsx
```diff
@@ -615,4 +618,212 @@ describe('FilterFacetsComponent', () => {
...
+          expect(ibmFacetValue).toBeDefined();
+          fireEvent.click(organizationCategoryHeader);
+          ibmFacetValue = fieldFacetsComponent.queryByText('ibm (138993)');
+          expect(ibmFacetValue).toBe(null);
+        });
+      });
+
+      describe('when clearing facets', () => {
+        test('can select facet values within one category and clear them', () => {
+          const { fieldFacetsComponent, performSearchMock } = setupResult;
```
ColbyJohnIBM, line 740: Could you please add some comments in some of the longer tests?

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/CategoryFacetsGroup.tsx
```diff
@@ -0,0 +1,106 @@
...
+      expandedCategoriesCopy.splice(indexOfCategory, 1);
+      setExpandedCategories(expandedCategoriesCopy);
+    } else {
+      setExpandedCategories(expandedCategories.concat(`${facetsLabel}-${category}`));
+    }
+  };
+
+  return (
+    <>
+      {Object.entries(facetsByCategory[`${facetsLabel}`].categories).map((entity: any) => {
```
maniax89, line 84: what exactly is `entity`? is it an array of strings? array of objects? adding typing to this would help future devs

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
```diff
@@ -1,12 +1,15 @@
...
 
-export const buildAggregationQuery = (
-  configuration: DiscoveryV2.QueryTermAggregation[]
-): string => {
-  const aggregation = configuration.map(term => {
-    return term.count
-      ? 'term(' + term.field + ',count:' + term.count + ')'
-      : 'term(' + term.field + ')';
+export const buildAggregationQuery = (configuration: QueryAggregationWithName[]): string => {
+  const aggregation = configuration!.map(term => {
```
maniax89, line 4: i don't think you need `configuration!` (just remove the `!`) -> i would say this is in general dangerous to use `!` outside of tests

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
```diff
@@ -1,12 +1,15 @@
...
+    const termCount = term.count ? ',count:' + term.count : '';
+    const termName = term.name ? ',name:' + term.name : '';
+    let nestedTypeTermAgg = '';
+    if (term.field!.includes('enriched_') && term.field!.includes('entities.text')) {
```
maniax89, line 8: can we type the `QueryAggregationWithName` to make `field` required? that way we don't have `!` in production code

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
```diff
@@ -1,12 +1,15 @@
...
+      const topLevelTermEntityField = term.field!.split('.')[0];
+      nestedTypeTermAgg = `.term(${topLevelTermEntityField}.entities.type,count:1)`;
```
maniax89, line 10: having `count:1` makes me ask -> what happens if we do have more than 1 type for a category? for example if `Apple` is the entity key, couldn't there be two types, `Company` and `Food`? in that case, do we want to drop types by limiting it to `1`?

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
```diff
@@ -33,3 +41,25 @@ export interface CollectionItem {
   id: string;
   label: string;
 }
+
+export interface QueryAggregationWithName extends DiscoveryV2.QueryAggregation {
+  field?: string;
```
maniax89, line 46: why is `field` optional?

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
```diff
@@ -33,3 +41,25 @@ export interface CollectionItem {
...
+  count?: number;
+  name?: string;
+}
+
+export interface SelectableFieldFacetWithCategory {
+  key?: string;
```
maniax89, line 52: shouldn't `key` be required?

packages/discovery-styles/scss/components/search-facets/_search-facets.scss
```diff
@@ -8,6 +8,36 @@
   }
 }
 
+.#{$prefix}--search-facets {
+  .#{$prefix}--search-facet--category {
```
maniax89, line 12: not sure if this was copied, but if we are following BEM format, we should probably do something like `--search-facet__category` where `search-facet` is the block and `category` is the element. usually `--` indicates a modifier

maniax89:
@ColbyJohnIBM https://github.com/watson-developer-cloud/discovery-components/pull/115#issuecomment-613467933

> The filter panel counts do not always match the search results count when a filter is applied. We should file another issue in wire-plugin-tracker

^ are you sure this is on the WIRE side? i think i had brought this up before in that our facet counts may not match after making selections across multiple categories (although that might be different than what you are pointing out here). notably, we decided to allow users to expand selections so the matching results count won't match up.

ColbyJohnIBM:
In the case above, it was a single filter within a single facet and the counts were still off. I would guess that in that simplest case, the counts should match?

It appeared that the issue may be related to the counts being case-sensitive, and the search results being case-insensitive, but that's just a hunch based on the results that were coming back.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/CategoryFacetsGroup.tsx
kabeaty: Thanks for the catch, I meant to go back and type this. Updated now 👍

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
kabeaty: I am limiting to 1 here so that the facet value only shows up underneath 1 category within the whole facet. I'm not sure about the ux of having a facet value show up twice to be selected within the same facet. Would it be selectable beneath each category grouping separately? Would its selection in one category select it in another?

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
kabeaty: The issue I keep running into here is that `DiscoveryV2.QueryAggregation` doesn't have `field` required (although `DiscoveryV2.QueryTermAggregation`, etc., which inherit from it do), but it's the type that is returning in the `QueryResponse` in the search client in `DiscoverySearch`. Therefore, there are errors in `DiscoverySearch` if it's not optional here between getting the response and passing it off to `buildAggregationQuery`. I haven't come up with a better solution here so far, but I'll keep thinking...

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
kabeaty: I was matching it to `SelectableDynamicFacets` since that is the similar selectable facet interface also used in `CollapsibleFacetsGroup`, but in this case, it is always required, so updating to required here 👍

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
kabeaty: If I can figure out a better way around the issue with `DiscoveryV2.QueryAggregation` not having field required but that being in the response from the search client, then yes. I agree that would be better.

packages/discovery-react-components/src/components/SearchFacets/components/__tests__/FieldFacets.test.tsx
kabeaty: Yes, done, thanks!

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
maniax89: are you able to use a type guard to enforce it to be a `QueryTermAggregation` instead of `QueryAggregation`?

packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
maniax89: maybe take a look at type guards? that should enforce a type as long as `QueryTermAggregation` extends `QueryAggregation` appropriately -> https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

kabeaty: Ah nice, thank you!

kabeaty:
> Verified it works locally and LGTM. One small note:
> 
> When tabbing through, I noticed that there is nothing to indicate when the facet chevron is focused on but other chevrons on the page have a focus state. Do we want to add one?
> Screen Shot 2020-04-13 at 1 44 33 PM

Good point. I took a look into this, and it looks like the issue is styles that are set in Tooling for buttons in search facets that we should update so that the focus state will be as expected there (update where border is being set to none for buttons there). I will add a note to the issue that this should be done when updating Tooling to get in these changes.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/CollapsibleFacetsGroup.tsx
```diff
@@ -70,8 +83,34 @@ export const CollapsibleFacetsGroup: FC<CollapsibleFacetsGroupProps> = ({
     setIsCollapsed(!isCollapsed);
   };
 
+  let facetsByCategory: FieldFacetsByCategory = {};
+  if (hasCategories) {
+    facetsByCategory[`${facetsLabel}`] = { categories: {} };
+    if (isSelectableQueryTermAggregationResult(facets)) {
+      facets.map(result => {
+        const resultType = result!.aggregations![0].results![0].key;
+        if (resultType in facetsByCategory[`${facetsLabel}`].categories) {
```
maniax89, line 92: i feel like this if-else can be DRY-ed up a bit, but we can leave that for a later PR

PR #123, opened 2020-04-10
fix: footer title field length expands parent component too much

JuanNicolasGarcia:

This fix prevents the footer title length from expanding its parent so much where it push the Improvement and Customize Panel component off the screen in tooling.

This is a companion PR with (https://github.ibm.com/Watson-Discovery/discovery-tooling/pull/7397).  To test this code, checkout this PR in components and the other PR in tooling, and use `yarn link-components`.

To verify the change works, load the sample project and search `IBM`.  See the search results.  Inspect the footer title of the first result.  Manually edit the text information with the footer tag to something very long.  See that the footer title will have an `...` at the end and does not push the Improvement and Customize Panel off the screen.

Before:

8cd6a500-5cb3-11ea-96ca-30f68119ca62

After:

Screen Shot 2020-04-10 at 10 12 49 AM


JuanNicolasGarcia:
Issue: https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1236

JuanNicolasGarcia:
Will update description when other PR in tooling is ready.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -119,7 +119,11 @@ export const Result: React.FunctionComponent<ResultProps> = ({
   const hasText = displayedText && !showTablesOnlyResults;
   const emptyResultContent = !(hasText || tableHtml);
 
-  const title = getDocumentTitle(result, resultTitleField);
+  let title = getDocumentTitle(result, resultTitleField);
+
+  if (Array.isArray(title)) {
```
H-Fin, line 124: Since its not obvious, maybe a brief comment explaining why this is needed might be nice.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -119,7 +119,11 @@ export const Result: React.FunctionComponent<ResultProps> = ({
...
+    title = title.join(', ');
```
maniax89, line 125: this doesn't make sense - why would we be joining the titles with commas? shouldn't we just select the first element? commas are not i18n-friendly for other languages

JuanNicolasGarcia: Honestly, I think just showing the first element is the best choice, but I was never really given any clear answer on what should be displayed when I asked about this.  Granted this is very much an edge case where someone would deliberately annotate several title fields and then search by the title field.

I can ask again to hopefully get a clear answer.

jhpedemonte: I agree with the maniac: don't do this; just select the first one.

packages/discovery-styles/scss/components/search-results/_search-results.scss
```diff
@@ -175,6 +175,7 @@ $result-table-fade-offset: calc(#{$result-element-max-height} - #{$result-table-
 }
 
 .#{$prefix}--search-result__footer__title {
+  max-width: 75%;
```
jhpedemonte, line 178: Why is this necessary?

JuanNicolasGarcia: If the max width of the footer title is not constrained in any way, it will actually occupy almost all of the footer space leaving no room for the collection name on the right side.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
JuanNicolasGarcia: This was the conclusion we reached in standup, will add this change.

packages/discovery-styles/scss/components/search-results/_search-results.scss
JuanNicolasGarcia: Changed to `flex: 1` and added extra margin on right of footer title.

PR #127, opened 2020-04-15
fix: update cypress and related dependencies

H-Fin:
Upgrade Cypress (and related deps) to match tooling.

No

### DIFF
--- a/examples/discovery-search-app/cypress/integration/passages/spec.ts
+++ b/examples/discovery-search-app/cypress/integration/passages/spec.ts
@@ -58,7 +58,7 @@ describe('Passage Results', () => {
         cy.get('.bx--document-preview').should('exist');
         cy.get('.bx--document-preview')
           .contains(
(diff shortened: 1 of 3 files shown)


examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -58,7 +58,7 @@ describe('Passage Results', () => {
         cy.get('.bx--document-preview').should('exist');
         cy.get('.bx--document-preview')
           .contains(
-            'This is a document.\nThis result multiple passages, but you should only be able to see the first one. IBM if you can see this passage, something probably borked'
+            'This is a document. This result multiple passages, but you should only be able to see the first one. IBM if you can see this passage, something probably borked'
```
H-Fin, line 61: This is necessary for this test to pass now since the Cypress version bump means that now: 

> `cy.contains()` ignores invisible whitespaces

per the [Migration Guide](https://docs.cypress.io/guides/references/migration-guide.html#cy-contains-ignores-invisible-whitespaces)

PR #130, opened 2020-04-20
fix: buildAggregationQuery should work with dictionary (nested and filter) aggregations

kabeaty:
During Integration Testing for the grouping top entities issue, Laxmi caught that dictionary aggregations (`nested` and `filter`) were causing an error. This is because their configuration wasn't checked for in the `buildAggregationQuery`. I've updated to check for them and build the query according to their configuration(s) as well.

Link these changes with Tooling and run against a project that includes dictionary facets. For example, there are two set up on `weary-griffin` at this time in the Sample project, if they're still there during testing. See that the expected facets render without error when searching. Check network requests and confirm that what is expected to be sent/received from the `query` endpoint is.

no

no

### DIFF
--- a/packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
+++ b/packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
@@ -2,14 +2,36 @@ import { QueryAggregationWithName } from './searchFacetInterfaces';
 
 export const buildAggregationQuery = (configuration: QueryAggregationWithName[]): string => {
   const aggregation = configuration.map(term => {
-    const termCount = term.count ? ',count:' + term.count : '';
-    const termName = term.name ? ',name:' + term.name : '';
-    let nestedTypeTermAgg = '';
-    if (term.field.includes('enriched_') && term.field.includes('entities.text')) {
-      const topLevelTermEntityField = term.field.split('.')[0];
-      nestedTypeTermAgg = `.term(${topLevelTermEntityField}.entities.type,count:1)`;
+    if (term.type === 'term') {
+      const termCount = term.count ? ',count:' + term.count : '';
+      const termName = term.name ? ',name:' + term.name : '';
+      let nestedTypeTermAgg = '';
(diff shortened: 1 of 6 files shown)


packages/discovery-react-components/src/components/SearchFacets/utils/buildAggregationQuery.ts
```diff
@@ -2,14 +2,36 @@ import { QueryAggregationWithName } from './searchFacetInterfaces';
 
 export const buildAggregationQuery = (configuration: QueryAggregationWithName[]): string => {
   const aggregation = configuration.map(term => {
-    const termCount = term.count ? ',count:' + term.count : '';
-    const termName = term.name ? ',name:' + term.name : '';
-    let nestedTypeTermAgg = '';
-    if (term.field.includes('enriched_') && term.field.includes('entities.text')) {
-      const topLevelTermEntityField = term.field.split('.')[0];
-      nestedTypeTermAgg = `.term(${topLevelTermEntityField}.entities.type,count:1)`;
+    if (term.type === 'term' && term.field) {
```
anirudhaBapat, line 5: To make things readable `term` from line 4 can be destruct

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
```diff
@@ -48,15 +48,23 @@ export interface CollectionItem {
...
+  aggregations?: QueryAggregationWithName[];
+  matching_results?: number;
 }
 
 export const isQueryAggregationWithName = (
   aggregations?: (DiscoveryV2.QueryAggregation | QueryAggregationWithName)[]
 ): aggregations is QueryAggregationWithName[] => {
-  return (aggregations as QueryAggregationWithName[])[0].field !== undefined;
+  return (
+    (aggregations as QueryAggregationWithName[])[0].field !== undefined ||
```
anirudhaBapat, line 64: Can this also be destructed?

H-Fin:
I will wait on merging this until the fix for master build is in.

PR #132, opened 2020-04-29
fix: failing cypress test

H-Fin:
This fixes the two failing cypress tests in `collection_filter` that are preventing the current master from building. 

I had to change the alias name in both tests since a describe that wraps these was using the same alias name. Not sure how the original test ever worked.

run `yarn cypress` and see that all tests now pass

N/A

No

### DIFF
--- a/examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
+++ b/examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
@@ -70,9 +70,11 @@ describe('Collection Filter', () => {
 
         describe('and we click the clear selected collections button', () => {
           beforeEach(() => {
-            cy.route('POST', '**/query?version=2019-01-01', '@queryJSON').as('newPostQuery');
+            cy.route('POST', '**/query?version=2019-01-01', '@queryJSON').as(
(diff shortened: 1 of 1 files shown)


ColbyJohnIBM:
Nice catch! I agree that I don't know how it ever worked... :shrug:

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -70,9 +70,9 @@ describe('Collection Filter', () => {
 
         describe('and we click the clear selected collections button', () => {
           beforeEach(() => {
-            cy.route('POST', '**/query?version=2019-01-01', '@queryJSON').as('postQuery');
+            cy.route('POST', '**/query?version=2019-01-01', '@queryJSON').as('newPostQuery');
```
noah-eigenfeld, line 73: Naming nitpick: for disco components, we tend to name mocked routes as `{type of request}{request endpoint}{specific name that differentials the request or response}`. So, maybe something like `postQueryClearedSelections` in this case?

noah-eigenfeld:
Ran the cypress tests locally, and they looked good!

PR #136, opened 2020-05-14
fix: search facet categories bug

H-Fin:

https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1544

Fixes a bug found during integration testing where facets will not load in certain double nested situations when the field name is `enriched_text.entities.text`. This is because these dictionary facets were being read as having categories when they did not.

~~Run this against `weary-griffin` Sample Project right now and see that facets will load. See that they currently not without this.~~

If that is gone, the same situation is now reflected as a test, see that tests pass. Remove the change in `FieldFacets.tsx` and see that the test will fail to load the facet group.

Tests have been updated.

No


JuanNicolasGarcia:
Wasn't able to recreate the test environment in `weary-griffin`.  Didn't see a difference when loading this branch and master branch.  However I did follow the secondary steps and tested with the updated `FieldFacets.test`.  That seems to work as intended.

H-Fin:
> Wasn't able to recreate the test environment in `weary-griffin`. Didn't see a difference when loading this branch and master branch. However I did follow the secondary steps and tested with the updated `FieldFacets.test`. That seems to work as intended.

That is true, the case is no longer on `weary-griffin`. I can try to recreate one later on there later but the test case added here reflects the same situation.

packages/discovery-react-components/src/components/SearchFacets/components/FieldFacets.tsx
```diff
@@ -117,7 +117,9 @@ export const FieldFacets: FC<FieldFacetsProps> = ({
         }
 
         const hasCategories =
-          aggregation.field.includes('enriched_') && aggregation.field.includes('entities.text');
+          aggregation.field.includes('enriched_') &&
+          aggregation.field.includes('entities.text') &&
+          aggregation.results![0].aggregations !== undefined;
```
NickMele, line 122: Can we be positive that `results` exists? Seems like it would be better to do optional chaining here if we can. 

```suggestion
          aggregation.results?.[0]?.aggregations !== undefined;
```

NickMele: Even if we cannot, I dont think i am comfortable with using the `!`. We have seen too many times that the API can return unexpected results.

H-Fin: Agreed. I used `get` instead since optional chaining is not available on our current version of `TypeScript`. This is ready for a re-review.

jhpedemonte: It isn't? I thought I fixed this a while back: https://github.com/watson-developer-cloud/discovery-components/pull/101

H-Fin: !!! Looking at that, and it looks like that PR never got merged into `master`. It went into `develop` and I'm not seeing those changes reflected in `master`.

jhpedemonte: Huh, did I accidentally set the base to `develop`? Well, color me embarrassed! I'll put up another PR.

jhpedemonte: Looks like we switched from `develop` to `master` on Mar 26, same day that PR went in. Must have gotten lost in the shuffle.

H-Fin: I was thinking it was probably something like that.

jhpedemonte: => https://github.com/watson-developer-cloud/discovery-components/pull/139

PR #138, opened 2020-05-16
fix: category facet not showing correct number

H-Fin:
Fixes a one off error inside the CategoryFacets so that they show up to 5 facets when `show more` is added. It was previously showing 4 and was inconsistent with the other facets.

Issue: https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1545

- Run storybook **without** the change in `CategoryFacets.tsx`
    - See that the CategoryFacet `Locations` and `Quantity` only have 4 facets
    - See `Products`, `Machine Learning`, and `Dynamic Facets` have 5 facets
    - See that the added test fails

- Add back the change and run storybook
    - See that both `Locations` and `Quantity` have 5 facets
    - See that the added test passes

Tests have been updated

No


JuanNicolasGarcia:
Tested and looks like things work.  Pretty straightforward change.

PR #141, opened 2020-06-02
feat: script that starts client AND server for example app

noah-eigenfeld:
In our manual setup instructions, we never tell the user that they now have to start not only the client but the server as well. This changes the name of the start client script to `start:client`, and adds a new `start` script that runs both the server and the client.

It also fixes I typo I spotted in the Windows Only section of our readme.

1. Run `yarn workspace discovery-search-app start:client` and make sure that it starts only the client at localhost:3000.
2. Run `yarn workspace discovery-search-app start` and make sure that it starts both the server (localhost:4000) and the client (localhost:3000).
N/A
No


README.md
```diff
@@ -180,125 +180,129 @@ If you don't have a React application already, start with [create react app](htt
...
-
-  const App = () => {
-    // see more detailed searchClient example above for `searchClient` variable
-    return (
-      <DiscoverySearch searchClient={searchClient} projectId={'REPLACE_ME'}>
-        <MyCustomComponent />
-      </DiscoverySearch>
-    );
-  };
+```jsx
```
noah-eigenfeld, line 183: I think these changes were made by my editor. Sigh...

I'll remove them.

examples/discovery-search-app/package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "build": "SKIP_PREFLIGHT_CHECK=true react-scripts build",
     "eject": "SKIP_PREFLIGHT_CHECK=true react-scripts eject",
-    "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
+    "start:client": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
+    "start": "yarn run server:run & yarn run start:client",
```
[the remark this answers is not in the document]
noah-eigenfeld: 💯agree. I was guessing this line would need to change, but hadn't looked up what the best practices were just yet. Thanks for the link!

examples/discovery-search-app/src/setupProxy.js
```diff
@@ -8,18 +8,18 @@ const cors = require('cors');
...
 
 if (clusterEnv.error) {
-  console.warn(clusterEnv.error);
+  throw new Error(clusterEnv.error);
 }
 
-const RELEASE_PATH = process.env.RELEASE_PATH;
-const BASE_URL = process.env.BASE_URL;
+const RELEASE_PATH = process.env.RELEASE_PATH || '/example/release/path';
+const BASE_URL = process.env.BASE_URL || 'http://example.com';
```
noah-eigenfeld, line 19: @maniax89 what are your thoughts on still giving these default values? If `.env.local` or `.server-env` don't exist, we'll throw exceptions. But that doesn't guarantee that any of the specific env vars exist within those files. Should we add checks for each one, or leave these default values in place?

examples/discovery-search-app/src/setupProxy.js
```diff
@@ -8,18 +8,18 @@ const cors = require('cors');
 const bodyParser = require('body-parser');
 
 if (serverEnv.error) {
-  console.warn(serverEnv.error);
+  throw new Error(serverEnv.error);
```
noah-eigenfeld, line 11: Noticed that the travis builds are failing here, since they must not have a `.server-env` file. Not sure what the best solution is.

examples/discovery-search-app/src/setupProxy.js
maniax89: I think it makes sense to check each one if the authenticator doesn't already

examples/discovery-search-app/src/setupProxy.js
maniax89: I would change the cypress tests to not start the server

README.md
```diff
@@ -111,7 +111,7 @@ yarn workspace discovery-search-app run start
 
    #### Windows Only
 
-   On Windows, the default `SASS_PATH` environmnt variable must be updated. Append the following to the `.env.local` file:
+   On Windows, the default `SASS_PATH` environment variable must be updated. Append the following to the `.env.local` file:
```
JuanNicolasGarcia, line 114: Nice

package.json
```diff
@@ -51,7 +51,8 @@
     "carbon-components": "^10.6.0",
     "carbon-components-react": "^7.7.0",
     "carbon-icons": "^7.0.7",
-    "cross-env": "^7.0.2",
+    "concurrently": "^5.2.0",
+    "cross-env": "^5.1.4",
```
JuanNicolasGarcia, line 55: Why did we revert to an older version of `cross-env`?  Does it have to with `concurrently`?

noah-eigenfeld: That's a good question. I'm guessing that we upped `cross-env` versions while this pr was in the wings, then maybe I crossed out the wrong one during a conflict resolution. I'll try changing it back to 7 and see if anything breaks...

noah-eigenfeld: Everything looked alright to me

examples/discovery-search-app/package.json
```diff
@@ -9,9 +9,10 @@
   "scripts": {
     "build": "cross-env SKIP_PREFLIGHT_CHECK=true react-scripts build",
     "eject": "cross-env SKIP_PREFLIGHT_CHECK=true react-scripts eject",
-    "start": "cross-env SKIP_PREFLIGHT_CHECK=true react-scripts start",
+    "start:client": "cross-env SKIP_PREFLIGHT_CHECK=true react-scripts start",
+    "start": "concurrently -n 'SERVER,CLIENT' -c 'blue,green' \"yarn run server:run\" \"yarn run start:client\" -k",
```
maniax89, line 13: maybe we should name this `start:prod` and have it only run `yarn run server:run`? i'm not sure what the true benefit is to start the server during normal operation

examples/discovery-search-app/package.json
```diff
@@ -9,9 +9,10 @@
...
     "lint": "eslint --quiet '{src,cypress}/**/*.{js,jsx,ts,tsx}' './*.{js,jsx,ts,tsx}'",
-    "test:e2e": "cross-env BROWSER=none CYPRESS_baseUrl=http://localhost:4000 start-server-and-test server:run http://localhost:4000 'cypress run'",
+    "test:e2e": "cross-env BROWSER=none CYPRESS_baseUrl=http://localhost:3000 start-server-and-test start:client http://localhost:3000 'cypress run'",
```
maniax89, line 15: this is good, but if you take the suggestion above, it would just be `start` instead of `start:client`

examples/discovery-search-app/src/setupProxy.js
maniax89: can we throw an error if `RELEASE_PATH` or `BASE_URL` are undefined along with an action the user to take? (like "please make sure you have defined `BASE_URL` in the `.server-env` file before attempting this command again")

noah-eigenfeld: I want to say that I tried, but it was causing additional problems, because `setupProxy` is run for both server and client startup (and the env vars are only present in server startup). I removed those changes, since the priority with this issue was to get the setup instructions in a good state. If we hear from customers that that part is unclear, I'm willing to return to this, or I could spin up a smaller issue to deal with it.

maniax89: yes that makes sense. probably some additional checks are needed when starting the client to ensure those variables are present since like you said both server and client rely on the same variables

examples/discovery-search-app/package.json
noah-eigenfeld: If the only thing this command does is run `yarn run server:run`, then I don't see why we need it at all.

examples/discovery-search-app/src/setupProxy.js
noah-eigenfeld: Actually, if we add a cypress env var like you suggested elsewhere, we might be able to bypass that problem. I'm going to fiddle with it and see if that works.

PR #142, opened 2020-06-22
chore: add cross-env to scripts requiring env var declaration

noah-eigenfeld:
Cross-env allows us to define environment vars in a react script without having those scripts break on Windows. However, it looks like we're only using them for a handful of the scripts that define environment vars in this way. This PR adds `cross-env` to any script that uses the pattern `SOME_ENV_VAR=value run whatever-script`.

On a Mac and again on Windows, check that the updated scripts run without needing to be modified.
Shouldn't be necessary, as we already had cross-env partially implemented
No


yarn.lock
```diff
@@ -2921,7 +2921,7 @@
     "@hapi/hoek" "8.x.x"
 
 "@ibm-watson/discovery-react-components@file:packages/discovery-react-components":
-  version "1.2.0-beta.10"
+  version "1.2.0-beta.12"
```
jhpedemonte, line 2924: Why this change?

noah-eigenfeld: Happened when I tried to `yarn add cross-env`, not realizing that we already had it installed. I can undo that if we want, but it didn't seem to hurt anything, so I just left it in.

jhpedemonte: Just realized it has `file:`. So this must have been added when linking to local source of the components. I assume this won't have any impact in production, so fine as is.

PR #143, opened 2020-06-26
fix: adds track event hooks

JuanNicolasGarcia:

Add track event hooks to call track event analytics in tooling.

Adds onChange callbacks to these components:

- SearchFacets (passes back synthetic event) 
- SearchInput (passes back search value `string`, value entered into search bar)
- SearchResults (passes back search value `string`  value of spelling suggestion)
- ResultsPagination (passes back object with page number and page size)

Requires `yarn link-components` command in tooling.

Companion Tooling PR: https://github.ibm.com/Watson-Discovery/discovery-tooling/pull/7655

The tooling PR has console logs with the event callback information, to verify the onChange callbacks are working.

Issue: https://github.ibm.com/Watson-Discovery/the-tooling-core-tracker/issues/1388


packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -49,6 +49,17 @@ interface SearchFacetsProps {
    * Number of facet terms to show when list is collapsed
    */
   collapsedFacetsCount?: number;
+  /**
+   * Used for analytics tracking
+   */
+  trackEventFacetsSelect?: (payload: {
```
maniax89, line 55: I don't think we want to expose properties related to tracking. What we can do is expose custom user interactions (like onClick or onKeyUp, etc) then reuse those props to add tracking events in tooling

JuanNicolasGarcia: @maniax89 okay, I understand not passing the needed properties up to the tooling track event function that exists there.  But I feel like I'm already doing what you're saying, I'm passing callback functions down into components that trigger the track event function call in tooling.  I'm probably misinterpreting what you're saying, but could you clarify a bit more or perhaps provide some psuedo example?  That would be helpful.

maniax89: its more about naming and reusability more than anything. once we commit to these properties, changing them requires a breaking change. 

we could just have a prop exposed called `onChange` that gets called with the native input events from checkboxes and radio buttons get triggered. that way users have access to all the same properties they would normally when invoking `onChange` events - see how carbon exposes their `onChange` event https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Checkbox/Checkbox.js#L45
```
onChange?: (e: SyntheticEvent<HTMLInputElement>) => void
```

then if that function is present, call it when triggering change events on the checkbox in https://github.com/watson-developer-cloud/discovery-components/blob/master/packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx#L67
```typescript
  const handleOnChange = (
    checked: boolean,
    _id: string,
    event: SyntheticEvent<HTMLInputElement>
  ): void => {
    if (onChange) {
      onChange(event);
    }
   // rest of handling code
  };
```

you will have to change some of the currently internal `onChange` symbols (they shouldn't have been named that anyway) to something more custom like `onFacetChange` ex:
```diff
<FieldFacets
            allFacets={allFieldFacets}
            showMatchingResults={showMatchingResults}
            onChange={onChange}
+.          onFacetChange={handleOnChange}
            collapsedFacetsCount={collapsedFacetsCount}
            messages={mergedMessages}
          />
```

jhpedemonte: I really like this idea ^^^. It's clean and generic (not just tied to analytics).

JuanNicolasGarcia: @maniax89 I've updated the PR with changes mostly to `SearchFacets`.  Wanted to know if this is the right direction I should be going in before attempting to clean things up.

I'm also in a situation where in the `SearchInput` component, I'm not entirely sure on how to proceed with that one.  It seems like there are two different onChange events that could be surfaced.  The `onChange` for input entry, and an `onKeyUp` for pressing the enter key to submit.  Each on their own are not that useful for segment tracking, but together, some logic could be cobbled together to make a useful track event.  Though I'm sure if we even want to do that.  Also those `onChange` events don't really give enough indication on whether the search query was typed in manually or selected using the autocomplete.

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -49,6 +49,10 @@ interface SearchFacetsProps {
    * Number of facet terms to show when list is collapsed
    */
   collapsedFacetsCount?: number;
+  /**
+   * Exposed onChange function for external use
```
maniax89, line 53: yes this is good - i would slightly modify the javadoc here to indicate which onChange event this corresponds to. something like

> custom handler invoked when any <input> element changes in this component

packages/discovery-react-components/src/components/SearchResults/SearchResults.tsx
```diff
@@ -77,6 +77,10 @@ export interface SearchResultsProps {
    * callback function from the component for sending document
    */
   onSelectResult?: (document: { document: DiscoveryV2.QueryResult }) => void | undefined;
+  /**
+   * Used for tracking analytics
+   */
+  trackEventSpelling?: (payload: { eventName: any; eventProps: { 'custom.term': string } }) => void;
```
maniax89, line 83: since this is a complex compound component, i could see this being called `onSpellingClick?: (e: SyntheticEvent<EventTarget>)`

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -187,7 +195,8 @@ const SearchFacets: FC<SearchFacetsProps> = ({
           <FieldFacets
             allFacets={allFieldFacets}
             showMatchingResults={showMatchingResults}
-            onChange={handleOnChange}
+            onChange={onChange}
+            onSearchFacetsChange={handleOnChange}
```
noah-eigenfeld, line 199: I'd recommend changing the name of `handleOnChange` to `handleOnSearchFacetsChange`, since I think that matches how the existing `onChange` has been renamed. Maybe we should even make the new `onChange` clearer in its naming, since we have two similar functions.

packages/discovery-react-components/src/components/SearchFacets/components/DynamicFacets.tsx
```diff
@@ -28,14 +28,19 @@ interface DynamicFacetsProps {
   /**
    * Callback to handle changes in selected facets
    */
-  onChange: (updatedFacet: Partial<SearchFilterFacets>) => void;
+  onSearchFacetsChange: (updatedFacet: Partial<SearchFilterFacets>) => void;
+  /**
+   * custom handler invoked when any input element changes in the SearchFacets component
+   */
+  onChange?: (e: SyntheticEvent<HTMLInputElement>) => void;
```
noah-eigenfeld, line 35: See my naming comment on `SearchFacets.tsx`, as it applies here as well.

packages/discovery-react-components/src/components/SearchFacets/components/DynamicFacets.tsx
```diff
@@ -71,7 +76,8 @@ export const DynamicFacets: FC<DynamicFacetsProps> = ({
       facets={dynamicFacets}
       showMatchingResults={showMatchingResults}
       facetsTextField="text"
-      onChange={handleOnChange}
+      onFieldFacetsChange={handleOnChange}
```
noah-eigenfeld, line 79: I also recommend renaming the `handleOnChange` here

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/CategoryFacetsGroup.tsx
```diff
@@ -29,7 +29,7 @@ interface CategoryFacetsGroupProps {
   /**
    * Callback to handle changes in selected facets
    */
-  onChange: (selectedFacets: SelectedFacet[]) => void;
+  onCollapsibleFacetsGroup: (selectedFacets: SelectedFacet[]) => void;
```
noah-eigenfeld, line 32: Following the naming convention you're using in other files, wouldn't this be `onCollapsibleFacetsGroupChange`?

packages/discovery-react-components/src/components/SearchInput/SearchInput.tsx
```diff
@@ -122,7 +127,8 @@ const SearchInput: FC<SearchInputProps> = ({
     const prefix = valueArray.pop();
     const completionValue = !!completions ? completions[i] : prefix;
     valueArray.push(completionValue || '');
-    setValue(`${valueArray.join(splitSearchQuerySelector)}${splitSearchQuerySelector}`);
+    const actualValue = `${valueArray.join(splitSearchQuerySelector)}${splitSearchQuerySelector}`;
```
noah-eigenfeld, line 130: What is the purpose of defining this variable instead of just passing all of this into the `setValue` function?

JuanNicolasGarcia: Hmm, this seems to be left over changes that I forgot to undo when redoing some work with this component.  I was using the `setValue` twice before, so it made sense then to make a `const`.  Will change back, since it's not necessary anymore.

packages/discovery-react-components/src/components/SearchFacets/components/DynamicFacets.tsx
```diff
@@ -71,7 +76,8 @@ export const DynamicFacets: FC<DynamicFacetsProps> = ({
       facets={dynamicFacets}
       showMatchingResults={showMatchingResults}
       facetsTextField="text"
-      onChange={handleOnChange}
+      onCollapsibleFacetsGroupChange={handleOnChange}
```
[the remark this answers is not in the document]
JuanNicolasGarcia: Love those long `const` names.

packages/discovery-react-components/src/components/SearchResults/components/SpellingSuggestion/SpellingSuggestion.tsx
```diff
@@ -10,9 +10,16 @@ interface SpellingSuggestionProps {
    * Message prefix used when displaying spelling suggestion
    */
   spellingSuggestionPrefix?: string;
+  /**
+   * Used for tracking analytics
```
maniax89, line 14: i would update this to match the javadoc you have in the other ones

JuanNicolasGarcia: Must have missed that one. Updated in latest commit.

PR #144, opened 2020-06-30
fix: updating rendering of html to add role presentation to tables

NickMele:
Table results are unable to be read by screen readers. For now, we are going to mark them as "presentation" so that the screen readers don't incorrectly them. We will follow up with a better solution that attempts to add headers to the data if they are available.

1. Run linked to tooling against cloud
2. Select a project with table results
3. When the results are returned in the search, and have a table in them run the [IBM Equal Accessibility Checker](https://chrome.google.com/webstore/detail/ibm-equal-access-accessib/lkcagbfjnkomcinoddgooolagloogehp) to verify it doesn't read the tables

No

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect, useState, useRef } from 'react';
+import React, { FC, useEffect, useState, useRef, useLayoutEffect } from 'react';
 import { settings } from 'carbon-components';
 import { QueryResult, QueryResultPassage, QueryTableResult } from 'ibm-watson/discovery/v2';
 import DOMPurify from 'dompurify';
@@ -60,7 +60,7 @@ export const HtmlView: FC<Props> = ({
   const [processedDoc, setProcessedDoc] = useState<ProcessedDoc | null>(null);
   const [highlightLocations, setHighlightLocations] = useState<Location[]>([]);
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     DOMPurify.addHook('afterSanitizeAttributes', function(node) {
       if (node.tagName === 'TABLE') {
         node.setAttribute('role', 'presentation');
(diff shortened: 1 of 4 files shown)


packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
```diff
@@ -60,6 +60,18 @@ export const HtmlView: FC<Props> = ({
   const [processedDoc, setProcessedDoc] = useState<ProcessedDoc | null>(null);
   const [highlightLocations, setHighlightLocations] = useState<Location[]>([]);
 
+  useEffect(() => {
```
maniax89, line 63: do you want to use https://reactjs.org/docs/hooks-reference.html#uselayouteffect here? i suppose `DOMPurify` handles rerendering just fine, but i thought `useLayoutEffect` was used exactly for this purpose (after rendering but before painting to make last minute modifications)

JuanNicolasGarcia:
Verified in the IBM Equal Accessibility Checker that tables do not show up

packages/discovery-react-components/src/components/DocumentPreview/components/HtmlView/HtmlView.tsx
NickMele: Good point, i will make that change. Thanks!

packages/discovery-react-components/src/components/SearchResults/components/ResultElement/ResultElement.tsx
```diff
@@ -71,9 +76,27 @@ export const ResultElement: React.FunctionComponent<ResultElementProps> = ({
...
+        }
+      });
+    }
+
+    return () => {
+      if (elementType === 'table') {
+        DOMPurify.removeHook('afterSanitizeAttributes');
+      }
+    };
+  }, [elementType]);
```
jhpedemonte, line 98: 1. This is duplicated in multiple places. Better to create a custom hook with the main logic.
2. This should also have a good comment on why the code is needed, what it is doing.
3. Why do we need the `removeHook`?

PR #146, opened 2020-08-05
refactor: remove global carbon components from search component sass files

hankthewhale:
This change removes carbon `import`s that are within each search component's `scss` files to remove the chance of duplicate imports that lead to longer build times.

___

#### The problem

1. Consumer app imports base carbon styles for its needs

**styles.scss**

```css
// Carbon global styles
@import 'carbon-components/scss/globals/scss/_css--font-face.scss';
@import 'carbon-components/scss/globals/scss/_colors.scss';
@import 'carbon-components/scss/globals/scss/_vars.scss';
@import 'carbon-components/scss/globals/scss/_mixins.scss';
@import 'carbon-components/scss/components/ui-shell/_variables.scss';
@import 'carbon-components/scss/components/ui-shell/_content.scss';
@import 'carbon-components/scss/components/ui-shell/_functions.scss';
@import 'carbon-components/scss/components/ui-shell/_header.scss';
@import '~carbon-components/scss/components/search/search';
@import '~carbon-components/scss/components/list-box/list-box';
```

2. Consumer app then imports search component styles

**styles.scss**

```css
// Carbon global styles
@import 'carbon-components/scss/globals/scss/_css--font-face.scss';
@import 'carbon-components/scss/globals/scss/_colors.scss';
@import 'carbon-components/scss/globals/scss/_vars.scss';
@import 'carbon-components/scss/globals/scss/_mixins.scss';
@import 'carbon-components/scss/components/ui-shell/_variables.scss';
@import 'carbon-components/scss/components/ui-shell/_content.scss';
@import 'carbon-components/scss/components/ui-shell/_functions.scss';
@import 'carbon-components/scss/components/ui-shell/_header.scss';
@import 'carbon-components/scss/components/search/search';
@import 'carbon-components/scss/components/list-box/list-box';

// Discovery component styles
@import '@ibm-watson/discovery-styles/scss/index.scss';

```

3. However, the discovery component index.scss file _also_ imports carbon global styles

**@ibm-watson/discovery-styles/scss/index.scss**

```css
@import '~carbon-components/scss/globals/scss/_css--font-face.scss';
@import '~carbon-components/scss/globals/scss/_typography.scss';
@import '~carbon-components/scss/globals/scss/_colors.scss';
@import '~carbon-components/scss/globals/scss/_layout.scss';
```

Additionally, each discovery component scss file imports the base carbon styles required for the component.

**@ibm-watson/discovery-styles/scss/components/search-input/_search-input.scss**

```css
@import '~carbon-components/scss/components/search/search';
@import '~carbon-components/scss/components/list-box/list-box';
```

Because the consumer app is already importing the base carbon styles required for its usage of carbon, anything that is also imported via discovery-components will be imported again, leading to longer build times. 

___

#### The proposed solution

Follow suite with carbon-react and stay "hands off" on base carbon style imports, but still provide the discovery component styles for consumption by the consumer app.

In this change, I'm removing all global carbon style imports from the discovery components and placing them into a `global-carbon-styles` file. This allows us to do the following:
1. Prevents duplicate imports if the consumer app is already importing this files.
2. Provides an easy way for the consumer app to import these global files if they do want to rely on the discovery-components to provide the global styles needed for these components. This would serve the use case where the consumer app is only interested in using the discovery components and doesn't want to have to figure out which imports are required from carbon while also serving the use case of apps that are using a mixed back of core carbon components plus add-ons.

Basically, instead of the consumer app doing:

```css
@import 'carbon-components/scss/globals/scss/_css--font-face.scss';
@import 'carbon-components/scss/globals/scss/_colors.scss';
@import 'carbon-components/scss/globals/scss/_vars.scss';
@import 'carbon-components/scss/globals/scss/_mixins.scss';
@import 'carbon-components/scss/components/ui-shell/_variables.scss';
@import 'carbon-components/scss/components/ui-shell/_content.scss';
@import 'carbon-components/scss/components/ui-shell/_functions.scss';
@import 'carbon-components/scss/components/ui-shell/_header.scss';
@import 'carbon-components/scss/components/search/search';
@import 'carbon-components/scss/components/list-box/list-box';

@import '@ibm-watson/discovery-styles/scss/index.scss';
```

it can instead do 

```
@import '@ibm-watson/discovery-styles/scss/carbon-components.scss';
@import '@ibm-watson/discovery-styles/scss/discovery-components.scss';
```

3. We can easily import both of these files into story book to continue having all of the carbon base styles available during package development.

Yes. Here is the sample app still running with the correct styles:

Screen Shot 2020-08-10 at 2 57 58 PM

Yes

I don't believe so. Because we still export the `index.scss` file which includes everything the same as before, this should be purely an additive change.


packages/discovery-styles/scss/global-carbon-styles.scss
```diff
@@ -0,0 +1,14 @@
...
+
+// Global styles
+@import '~carbon-components/scss/globals/scss/_css--font-face.scss';
+@import '~carbon-components/scss/globals/scss/_typography.scss';
+@import '~carbon-components/scss/globals/scss/_colors.scss';
+@import '~carbon-components/scss/globals/scss/_layout.scss';
+
+// Search Input
+@import '~carbon-components/scss/components/search/search';
+@import '~carbon-components/scss/components/list-box/list-box';
```
hankthewhale, line 14: This file is completely optional for consumer apps. They can choose to import this file along with the base discovery components scss file if they don't plan on mixing in any additional carbon components. I imagine this would be more likely used by users who are quickly building a sample search app. Perhaps as their project grows, they would abandon importing this file in favor of importing from carbon directly. Bottom line is that this allows to provide a decent developer experience for those who want a quick path to get something going while also saying "hey, you don't need to use this file if you have a more complex setup to manage."

packages/discovery-styles/scss/global-carbon-styles.scss
```diff
@@ -0,0 +1,14 @@
+// You only need to import this file if your app is not already directly importing these carbon files
```
hankthewhale, line 1: If we like this approach, I'd write up a better statement here that links to docs including this: https://github.com/carbon-design-system/carbon/blob/master/docs/guides/sass.md#optimizing-your-sass-builds

packages/discovery-styles/scss/components/search-input/_search-input.scss
```diff
@@ -1,7 +1,3 @@
-// Component styles
-@import '~carbon-components/scss/components/search/search';
```
maniax89, line 2: does `@use` work here? https://css-tricks.com/introducing-sass-modules/#:~:text=The%20new%20%40use%20is%20similar,considered%20private%2C%20and%20not%20imported.

or can we at least leave a comment indicating which carbon styles this file depends on?

hankthewhale: It seems like it would work, but we would have to use `dart-sass` instead of `node-sass` which then requires anyone consuming this package to also use `dart-sass`. 

At this point in time, it doesn't appear that `@use` is on the `node-sass` roadmap: https://github.com/sass/node-sass/issues/2886#issuecomment-608280859

hankthewhale: @maniax89 would you prefer comments per file or references within this file: https://github.com/watson-developer-cloud/discovery-components/blob/62561bc372d6d19c4d1a91ce1e12414cac1a0137/packages/discovery-styles/scss/global-carbon-styles.scss#L12

maniax89: either or - i don't have a preference. just as long as it is consistent

hankthewhale: Seems like many components import the same base styles. For example

https://github.com/watson-developer-cloud/discovery-components/blob/master/packages/discovery-styles/scss/components/ci-document/_ci-document-filter.scss#L1-L3

and 

https://github.com/watson-developer-cloud/discovery-components/blob/master/packages/discovery-styles/scss/components/search-facets/_search-facets.scss#L1-L3

Thinking that I'll go the route of putting all of the comments in the global file and include a section of shared base carbon styles across components.

hankthewhale: Did that here: https://github.com/watson-developer-cloud/discovery-components/pull/146/files#diff-bb1e94d2444f11bfc036c8c60f806d24

hankthewhale:
Going to be doing that today before I move this out of draft. Also will be updating docs.

README.md
```diff
@@ -301,6 +302,45 @@ If you don't have a React application already, start with [create react app](htt
   export default MyCustomComponent;
 ```
 
+### Minimal css best practices
+
+The Discovery Components styles package can be consumed in three ways
+
+1. Wholesale SCSS
```
hankthewhale, line 309: Not married to the definitions here. Just wanted something to quickly describe what the method is. Maybe single import SCSS and partial SCSS imports instead?

README.md
```diff
@@ -301,6 +302,45 @@ If you don't have a React application already, start with [create react app](htt
   export default MyCustomComponent;
 ```
 
+### Minimal css best practices
```
jhpedemonte, line 305: How about:

```suggestion
### Optimize CSS
```

or at the very least:

```suggestion
### Minimal CSS best practices
```

hankthewhale: Did the first

PR #147, opened 2020-08-05
fix: add collections facets hook for event tracking

JuanNicolasGarcia:

This adds an event callback for toggling collection facets that can be used for event tracking.
Companion tooling PR: https://github.ibm.com/Watson-Discovery/discovery-tooling/pull/7655

Pull down the tooling PR branch and use `yarn link-components` to test changes.
In a project workspace (sample project is fine), perform any search.  Select a collection facet, and see in the console a list of selected collection facets.

### DIFF
--- a/packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
+++ b/packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
@@ -26,6 +26,10 @@ import { withErrorBoundary } from 'react-error-boundary';
 
 interface SearchFacetsProps {
   /**
+   * ID for the SearchInput
+   */
+  id?: string;
+  /**
    * Show list of collections as facets
    */
   showCollections?: boolean;
@@ -50,12 +54,14 @@ interface SearchFacetsProps {
    */
   collapsedFacetsCount?: number;
   /**
-   * custom handler invoked when any input element changes in the SearchFacets component
+   * Custom handler invoked when any input element changes in the SearchFacets component.
+   * Takes a synthethic event from an HTML Input Element or a string array.
    */
   onChange?: (e: SyntheticEvent<HTMLInputElement> | string[]) => void;
 }
 
 const SearchFacets: FC<SearchFacetsProps> = ({
+  id = '',
   showCollections = true,
   showDynamicFacets = true,
   showMatchingResults = false,
@@ -102,6 +108,10 @@ const SearchFacets: FC<SearchFacetsProps> = ({
     overrideComponentSettingsAggregations ||
     (componentSettings && componentSettings.aggregations) ||
     [];
+
+  if (id !== '') {
+    id = id.concat('--');
+  }
 
   useEffect(() => {
     fetchAggregations(searchParameters);
@@ -195,7 +205,7 @@ const SearchFacets: FC<SearchFacetsProps> = ({
         {hasSelection && (
           <Button
             className={`${settings.prefix}--search-facets__button-clear-all`}
-            id={'search-facets-button-clear-all'}
+            id={`${id}search-facets-button-clear-all`}
             kind="ghost"
             renderIcon={Close}
             size="small"
--- a/packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
+++ b/packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
@@ -52,7 +52,7 @@ interface SearchFacetsProps {
   /**
    * custom handler invoked when any input element changes in the SearchFacets component
    */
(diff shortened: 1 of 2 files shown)


packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/CollapsibleFacetsGroup.tsx
```diff
@@ -152,6 +155,7 @@ export const CollapsibleFacetsGroup: FC<CollapsibleFacetsGroupProps> = ({
           {facetsLabel}
           {shouldDisplayClearButton && (
             <ListBox.Selection
+              id="search-facets-clear-listbox"
```
maniax89, line 158: this isn't unique to the page as there can be multiple `CollapsibleFacetsGroup`s on the page. If you want to assign identifiers, i would make sure to define a prefix controlled at the top-level component (`SearchFacets`) so users can override it as needed

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -183,6 +195,7 @@ const SearchFacets: FC<SearchFacetsProps> = ({
         {hasSelection && (
           <Button
             className={`${settings.prefix}--search-facets__button-clear-all`}
+            id={'search-facets-button-clear-all'}
```
maniax89, line 198: same here with identifiers, although this may be unique, nothing excludes a user from doing multiple `<SearchFacets>` on a page. adding an `id` would make it invalid HTML when multiple of the same id reside on the page. you can use `data-testid` or something similar if this is being used for testing

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -52,7 +52,7 @@ interface SearchFacetsProps {
   /**
    * custom handler invoked when any input element changes in the SearchFacets component
    */
-  onChange?: (e: SyntheticEvent<HTMLInputElement>) => void;
+  onChange?: (e: SyntheticEvent<HTMLInputElement> | string[]) => void;
```
maniax89, line 55: i would explain more about what `string[]` is in the javadoc here. `SyntheticEvent<HTMLInputElement>` is self-documenting. `string[]` is not

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
JuanNicolasGarcia: @maniax89 for this case and the one mentioned above, would a `data-testid` be suitable for tracking purposes?  With the scenario you mentioned with multiple `SearchFacets` being called, tracking would still only be concerned if that type of button is clicked rather than which specific one.  Since tracking does not record which specific UI element triggered the event.  If not, what would be a suitable method for placing identifiers for these components?

maniax89: for any of the carbon components, i would just make sure it is prefixed with something that is user-controlled if necessary. you can have a top-level attribute like `clearAllButtonId` that defaults to `search-facets-button-clear-all`. that way when users consume it in their applications, they don't experience accessibility violations due to duplicate ids https://www.w3.org/TR/WCAG20-TECHS/failures#F77

JuanNicolasGarcia: So user-defined would be something like what is occurring in the `SearchInput` component?
https://github.com/watson-developer-cloud/discovery-components/blob/2f68225950583370a2f224066faba1e1feaf0ea7/packages/discovery-react-components/src/components/SearchInput/SearchInput.tsx#L30

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -25,6 +25,10 @@ import { FallbackComponent } from 'utils/FallbackComponent';
 import { withErrorBoundary } from 'react-error-boundary';
 
 interface SearchFacetsProps {
+  /**
+   * ID for the SearchInput
+   */
+  id?: string;
```
maniax89, line 31: the way this is being used, its more of an `idPrefix` or `idSuffix`. `id` is a bit generic for this property since its only getting applied to a sub-component rather than the root wrapper object.

i would either:
a) attach this id to the root level and have it default to something if not set
b) append `prefix` or `suffix` to the name of this attribute so that its usage matches the name

JuanNicolasGarcia: I'm going to go with an implementation similar to the `SearchInput` component, where the id prop passed in is attached to the root `div` of the component.  If nothing is passed in, a default is given.

PR #152, opened 2020-08-17
fix: more error handling

jhpedemonte:

Better error handling


packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -270,21 +270,22 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
+          if (isQueryAggregationWithName(response.result.aggregations)) {
+            const updatedAggregations = await fetchTypeForTopEntitiesAggregation(
+              response.result.aggregations,
+              searchParameters
+            );
+            setAggregationResults(updatedAggregations || null);
+          } else {
+            setAggregationResults(response.result.aggregations);
+          }
+        }
```
jhpedemonte, line 288: Switch to using `await` instead of Promise syntax, so errors are correctly bubbled up.

PR #158, opened 2020-10-07
fix: html highlighting

jhpedemonte:

Fix how highlighting is handled:
1. Fix issue where if offset had value of zero, highlight was ignored.
2. If background-color was applied, then highlights in HTML docs disappeared (z-index issue).

This also updates the CSS for highlights rects in `SimpleDocument` to match the implementation in `HtmlView`.

To verify fix (1):
1. Start storybook and load **DocumentPreview > passage highlighting**
2. Using React Dev Tools plugin, select the `HtmlView` component
3. In box at upper-right (labeled "props"), update the `highlight` property to have `end_offset` of 20 and a `start_offset` of 0.
4. Notice that highlight properly renders

To verify fix (2):
1. In same storybook, go to **Inspector > Elements** and select the node `<div class="bx--document-preview">`
2. Add style of `background-color: white`.
3. Highlight rects should still show.

Compare against deployed instance: https://watson-developer-cloud.github.io/discovery-components/storybook/?path=/story/documentpreview--passage-highlighting

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/Highlight/passages.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/Highlight/passages.tsx
@@ -37,7 +37,7 @@ export function getPassagePageInfo(
 ): ReadonlyArray<CellPage> | null {
   const { start_offset, end_offset, field } = passage;
 
-  if (!start_offset || !end_offset) {
+  if (typeof start_offset !== 'number' || typeof end_offset !== 'number') {
     return null;
   }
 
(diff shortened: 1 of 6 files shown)


packages/discovery-react-components/src/components/DocumentPreview/components/Highlight/passages.tsx
```diff
@@ -37,7 +37,7 @@ export function getPassagePageInfo(
 ): ReadonlyArray<CellPage> | null {
   const { start_offset, end_offset, field } = passage;
 
-  if (!start_offset || !end_offset) {
+  if (typeof start_offset !== 'number' || typeof end_offset !== 'number') {
```
ColbyJohnIBM, line 40: Nice catch. I'm sure this had caused bugs with `0` in the past.

ColbyJohnIBM (approved):
Looks good. 💯

PR #159, opened 2020-10-09
refactor: carbon spacing tokens

hankthewhale:
This PR updates tooling to use the correct carbon 10 vars. Luckily there was a 1:1 mapping, so this PR is a find and replace.

I decided to hold off on the layout migration for this PR since the file count is going to be large here.

References:
https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/globals/scss/migrate-to-10.x.md#_spacingscss

https://www.carbondesignsystem.com/guidelines/spacing/overview/#spacing-scale

All components should retain the same padding and margins as before.

No.

No

### DIFF
--- a/examples/discovery-search-app/src/app.scss
+++ b/examples/discovery-search-app/src/app.scss
@@ -35,7 +35,7 @@ h1 {
   background-color: $search-app-background-color;
   max-height: 1500px;
   max-width: 1200px;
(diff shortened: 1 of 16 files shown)


jhpedemonte (approved):
I do not like this new nomenclature. But ... comme ci comme ça.

kabeaty (approved):

PR #160, opened 2020-12-22
fix: yarn upgrade

maniax89:

`yarn audit` was reporting some vulnerabilities - now it doesn't

`0 vulnerabilities found - Packages audited: 3204`

if travis build passes, I think we are ok

N/A

N/A

### DIFF
--- a/examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
+++ b/examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
@@ -34,13 +34,13 @@ describe('Collection Filter', () => {
       });
 
       it('displays a list of the collections in the project', () => {
-        cy.get('#collection-facet-select__menu')
+        cy.get('.bx--list-box__menu')
           .contains('deadspin')
           .should('exist');
-        cy.get('#collection-facet-select__menu')
+        cy.get('.bx--list-box__menu')
           .contains('espn')
           .should('exist');
-        cy.get('#collection-facet-select__menu')
+        cy.get('.bx--list-box__menu')
           .contains('finnegans wake')
           .should('exist');
       });
@@ -126,7 +126,7 @@ describe('Collection Filter', () => {
 
         it('the collection filter dropdown disappears', () => {
           cy.get('#collection-facet-select').should('exist'); // keeps a white screen from passing the test
-          cy.get('#collection-facet-select__menu').should('not.exist');
+          cy.get('.bx--list-box__menu').should('not.be.visible');
         });
       });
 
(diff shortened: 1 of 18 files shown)


hankthewhale (approved):

NickMele (approved):

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
```diff
@@ -34,13 +34,13 @@ describe('Collection Filter', () => {
       });
 
       it('displays a list of the collections in the project', () => {
-        cy.get('#collection-facet-select__menu')
+        cy.get('.bx--list-box__menu')
```
NickMele, line 37: Why are you switching to classes here?

packages/discovery-react-components/package.json
```diff
@@ -15,7 +15,7 @@
     "circular": "madge --circular src/*",
     "build": "rollup -c",
     "eslint": "eslint --quiet '{src,.storybook}/**/*.{js,jsx,ts,tsx}'",
-    "lint": "yarn run eslint && yarn run circular",
+    "lint": "yarn run circular && yarn run eslint",
```
NickMele, line 18: Not asking to revert this, just curious if there was a reason these needed to be swapped.

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
maniax89: i think when i upgraded carbon accidentally it stopped passing the `id` parameter to the internal downshift components, thus this selector no longer existed. im open to other selectors, the other one being `#downshift-0-menu`

packages/discovery-react-components/package.json
maniax89: our husky config passes `--fix` along to `yarn lint`. `circular` must have been ignoring extra parameters before. now it gets upset. this allows `--fix` to properly be passed to `eslint` instead

examples/discovery-search-app/cypress/integration/collection_filter/spec.ts
NickMele: Ahh ok. I would prefer an id if possible. but if not since we don't control it, that is ok.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/__tests__/CIDocument.spec.tsx
```diff
@@ -209,11 +209,11 @@ describe('<CIDocument />', () => {
     it('displays the correct party information on the details pane', async () => {
       // Expects metadata pane to be hidden until selected
       const metadataPane = getByTestId('metadata-pane').parentElement!;
-      expect(metadataPane.getAttribute('aria-hidden')).toEqual('true');
+      expect(metadataPane).toHaveProperty('hidden', true);
```
maniax89, line 212: in order to use `expect(await screen.findByRole('tab', { name: 'Document', selected: false })).toBeVisible();` we need to upgrade `@testing-library/react` from `9` to `>=10`, which causes a bunch of other test failures

PR #165, opened 2021-01-13
fix: facet selection state lost after results previewed

kabeaty:
These changes address the issue that the facet selection state is lost after a result is selected to be previewed. This gets the aggregations using `useDataApi` too for fetching aggregations and state.

I also fixed the misalignment of facet name / checkbox within this PR, as it was a small change that was bugging me while I fixed this.

I'm removing `yarn audit` in this PR, but it will be added back in after vulnerabilities are fixed, in a way that isn't part of the normal build process and blocking PRs.

I added a Cypress test for this. It's also possible to link to Tooling and test previewing documents/tables and returning to the workspace there with facets selected.


packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -143,8 +146,15 @@ const SearchFacets: FC<SearchFacetsProps> = ({
...
 
-    if (searchParamsAggregationChanged || fetchState === 'init') {
+    if (
+      (searchParamsAggregationChanged || fetchState === 'init') &&
+      !isReturningFromResultPreview
+    ) {
       fetchData();
+    } else {
+      // we don't need to refetch data if returning to search results and facets from previewing a result
+      setIsReturningFromResultPreview(false);
```
maniax89, line 156: im struggling to understand why this is necessary - is `fetchState` being reset to `init` when clicking on a search result? is there any way we could move the initial `fetchState` outside of this component?

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -143,8 +146,15 @@ const SearchFacets: FC<SearchFacetsProps> = ({
       }
     }
 
-    if (searchParamsAggregationChanged || fetchState === 'init') {
+    if (
+      (searchParamsAggregationChanged || fetchState === 'init') &&
+      !isReturningFromResultPreview
```
NickMele, line 151: Is the main purpose of this just to prevent aggregations from being fetched again? To me having a state var like this feels a bit weird. Do we know which of the previous conditions were true causing these to refetch? Was `fetchState` set to `init`? or did the search params aggregation change?

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
kabeaty: Yes exactly--`fetchState` was being reset to `init` when previewing a search result and then returning to search results/facets. I moved the `fetchState` out of this component into `DiscoverySearch` instead and like this approach better, thanks! 👍

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
kabeaty: I refactored to move the `fetchState` up to `DiscoverySearch` instead. The main issue was `fetchState` being reset to `init` after document preview, yes 👍

packages/discovery-react-components/src/components/SearchFacets/utils/searchFacetInterfaces.ts
```diff
@@ -92,3 +92,10 @@ export interface FieldFacetsByCategoryEntity {
     facets: SelectableFieldFacetWithCategory[];
   };
 }
+
+export enum FetchAggregationStates {
```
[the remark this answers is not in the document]
kabeaty: I was thinking it made sense to have this definition here and then just import this type into `DiscoverySearch` since this is where the types related to the facets/aggregations are defined.

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -131,22 +130,21 @@ const SearchFacets: FC<SearchFacetsProps> = ({
     (componentSettings && componentSettings.aggregations) ||
     [];
 
-  const searchParamsAggregationChanged = useCompare(searchParameters.aggregation);
   useEffect(() => {
     async function fetchData() {
-      setFetchState('loading');
+      setFetchAggregationState(FetchAggregationStates.LOADING);
```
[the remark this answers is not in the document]
kabeaty: I think it makes sense to keep it in the `SearchFacets` component since this is specific to the state for the facets/aggregations.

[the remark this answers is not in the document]
maniax89: instead of setting the fetch aggregation state here, can we just convert this to using the `useDataApi` similar to how we do it for search results?  https://github.com/watson-developer-cloud/discovery-components/blob/master/packages/discovery-react-components/src/utils/useDataApi.ts#L239-L277

something like
```typescript
export const useGlobalAggregationsApi = (
  searchParameters: DiscoveryV2.QueryParams,
  searchClient: SearchClient
): [GlobalAggregationsResponseStore, GlobalAggregationsStoreActions] => {
  const {
    state: aggregationState,
    parameters: currentGlobalAggregationParameters,
    setParameters: setGlobalAggregationParameters,
    setData: setGlobalAggregationsResponse,
    setFetchToken
  } = useDataApi(searchParameters, null, searchClient.query, searchClient);

  return [
    {
      ...aggregationState,
      parameters: currentGlobalAggregationParameters
    },
    {
      setGlobalAggregationParameters,
      setGlobalAggregationsResponse,
    }
  ];
};
```

then use it like
```typescript
const [globalAggregationsResponseStore, { fetchGlobalAggregations, setGlobalAggregationsParameters }] = useGlobalAggregationsApi(
    { ...fetchAggregationsResponseStoreDefaults.parameters, projectId },
    searchClient
  );
```

make sure to add the `fetchGlobalAggregations` as one of the api methods in the `SearchContext` as well as `globalAggregationsResponseStore` as part of the `SearchContextApi`

then you can just check the fetch state in the SearchFacets component like
```
const { fetchGlobalAggregations } = useContext(SearchApi);
const { globalAggregationsResponseStore: { isLoading, isError, data }} = useContext(SearchContext);
```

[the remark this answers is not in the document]
kabeaty: Yes, good point and thank you for the details. I've updated to use the `useDataApi` too 👍

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -107,12 +109,13 @@ export interface SearchContextIFC {
...
   isResultsPaginationComponentHidden: boolean | undefined;
   fieldsStore: FieldsStore;
+  globalAggregationsResponseStore: GlobalAggregationsResponseStore;
 }
 
 export interface SearchApiIFC {
   performSearch: (searchParameters: DiscoveryV2.QueryParams, resetAggregations?: boolean) => void;
   fetchAutocompletions: (nlq: string) => Promise<void>;
-  fetchAggregations: (searchParameters: DiscoveryV2.QueryParams) => Promise<void>;
+  fetchGlobalAggregations: (searchParameters: DiscoveryV2.QueryParams) => Promise<void>;
```
maniax89, line 118: im wondering if its worth it to change the name of this api because technically it would be a breaking change to this components repo (i doubt anyone is actually using it). I can be convinced to keep it this way for the sake of clarity, but just think its good to weigh the following:
1. keep this named `fetchAggregations` but have the conflict in meaning within the codebase (really it refers to global aggregations)
2. change it to `fetchGlobalAggregations` and do not do a major version increase (could break a client unexpectedly)
3. change it to `fetchGlobalAggregations` and do a major version increase (requires a migration guide when we publish the major version, capturing this commit as such)
i'm open to any options, want to hear your opinion

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -197,6 +209,7 @@ const fieldsStoreDefaults: FieldsStore = {
 export const searchContextDefaults = {
   aggregationResults: null,
   searchResponseStore: searchResponseStoreDefaults,
+  globalAggregationsResponseStore: globalAggregationsResponseStoreDefaults,
```
maniax89, line 212: does this replace `aggregationResults` or is it an entirely different set of data? i can't quite remember unfortunately

maniax89 (approved):
nice work! this looks a lot cleaner now. with my comments responded to, approved!

.travis.yml
```diff
@@ -13,7 +13,6 @@ branches:
     - /^hotfix\/.*/
 script:
   - yarn lint
-  - yarn audit
```
NickMele, line 16: Why did you remove this?

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
kabeaty: Good point, thanks. I think in this case that I lean toward 1, as I agree it is probably not worth it for this potentially breaking change and I think the meaning is still clear without this name change. Updated.

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
kabeaty: Ah, yes, thanks. I meant to remove this. Updated without `aggregationResults` and also updated the `overrideAggregationResults` to work with the new `globalAggregationsResponseStore`.

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -120,33 +119,23 @@ const SearchFacets: FC<SearchFacetsProps> = ({
...
-        await fetchAggregations(searchParameters);
-        setFetchState('success');
-      } catch (error) {
-        setFetchState('error');
-      }
+      await fetchGlobalAggregations(searchParameters);
     }
 
-    if (searchParamsAggregationChanged || fetchState === 'init') {
+    if (!aggregations) {
```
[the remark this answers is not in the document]
kabeaty: As discussed, it doesn't seem like we actually need this check here to support any use case.

examples/discovery-search-app/cypress.json
```diff
@@ -2,5 +2,6 @@
   "baseUrl": "http://localhost:3000/",
   "supportFile": "cypress/support/index.ts",
   "defaultCommandTimeout": 30000,
-  "video": false
+  "video": false,
+  "videoCompression": false
```
NickMele, line 6: Why does video compression need to be off if video is off?

maniax89 (approved):

PR #166, opened 2021-01-15
fix: Upgrade watson sdk

maniax89:

`axios` had a vulnerability in one of our packages (`ibm-watson`) so we needed to upgrade that. Unfortunately since we had to go up a major version to get it (we were on `alpha`), we had breaking changes detailed in https://github.com/watson-developer-cloud/node-sdk/wiki/v5-Migration-Guide#discovery

- `returnFields` was renamed to `_return`, we added a deprecation warning to anyone who uses this library and still has `returnFields` but copies it over to `_return` while logging a warning.
- `prefix` became required on the `GetAutocompleteParameters` type

Also took the time to clean up some types for `useDataApi` utility and `DiscoverySearch` component to more closely align with reality without breaking any public APIs

`lerna run test`

TBD

no breaking change until next major version

### DIFF
--- a/examples/discovery-search-app/package.json
+++ b/examples/discovery-search-app/package.json
@@ -32,7 +32,7 @@
     "dotenv": "^8.1.0",
     "express": "^4.17.1",
     "http-proxy-middleware": "^0.20.0",
-    "ibm-watson": "5.2.0-alpha-1",
+    "ibm-watson": "^6.0.2",
     "prop-types": "^15.6.2",
     "react": "^16.4.1",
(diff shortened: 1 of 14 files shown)


kabeaty (approved):
Looks good! Should the test [here](https://github.com/watson-developer-cloud/discovery-components/blob/master/packages/discovery-react-components/src/utils/__tests__/useDataApi.test.tsx#L692) in `useDataApi.test.tsx` also be updated to use `_return`? Also, is there anywhere that we're already keeping track of things like this that support will be removed for in the next major release?

PR #170, opened 2021-01-25
fix: keep aggregationResults in SearchContext

kabeaty:
This updates to add `aggregationResults` back into the `SearchContext`. This keeps it in sync with the `globalAggregationsResponseStore`, which is replacing the `aggregationResults`. However, since this is a breaking change, we should keep supporting it until the next major version.

You can test against CM in Tooling. Use `const { aggregationResults } = useContext(SearchContext);` instead of grabbing the aggregation results from the global aggregations response store there and see that the aggregation results correctly load for a CM project. Test that without these updates, using `aggregationResults`, they don't.

Included a comment in the code. Should we document upcoming breaking changes anywhere else?


packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -332,6 +336,12 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
     overrideAutocompletionResults
   );
 
+  // Keep aggregationResults in sync with globalAggregationsResponseStore data
+  // to support it until deprecated in next major release
+  useDeepCompareEffect(() => {
+    setAggregationResults(globalAggregationsResponseStore.data);
```
maniax89, line 342: could you write a quick spec to validate this?

maniax89 (approved):

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
kabeaty: Yes, updated with a test

maniax89 (approved):

PR #181, opened 2021-02-04
fix: do not set count on search params for fetchAggregations

maniax89:

need to prevent setting the searchParameters for the search response store when doing aggregation fetches. this was setting the count to `0` due to defaults

`/runExampleApp.sh` and perform a bunch of searches (empty) and validate you continue to get results back (albeit slow ones at times)

No

No

### DIFF
--- a/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
+++ b/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
@@ -294,6 +294,8 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
       ) as DiscoveryV2.QueryParams;
       setSearchParameters(searchParameters);
 
+      // TODO proper fix is to have any caller of handleSearch that needs aggregations
+      // to call handleFetchAggregations instead
       if (resetAggregations) {
         const aggregationSearchParams = {
           ...searchParameters,
--- a/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
+++ b/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
@@ -289,6 +289,7 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
       backwardsCompatibleQueryParams: DiscoveryV2.QueryParams & { returnFields?: string[] },
       resetAggregations = true
     ): Promise<void> => {
+      let aggregationsFetched = false;
       const searchParameters = deprecateReturnFields(
         backwardsCompatibleQueryParams
       ) as DiscoveryV2.QueryParams;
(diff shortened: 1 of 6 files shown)


NickMele (approved):

jhpedemonte (commented):
Code looks good and fixes the issue.

I did notice one quirk: there is a `/query` request with an empty object payload. To reproduce:
- load a project workspace (with search bar)
- clear Network inspector and set to only show XHR
- do a blank search
- notice 2 query requests. 2nd one has empty object payload

jhpedemonte (approved):

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
```diff
@@ -219,7 +219,7 @@ const SearchFacets: FC<SearchFacetsProps> = ({
     performSearch({ ...searchParameters, collectionIds: [], offset: 0, filter: '' }, false);
   };
 
-  if (isLoading) {
+  if (isLoading && !aggregations) {
```
maniax89, line 222: if we don't want this change i can remove it - was just somewhat jarring to see the aggregations disappearing after every search

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -287,31 +289,22 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
-            ...searchParameters,
-            ...aggregationQueryDefaults,
-            filter: ''
-          },
-          async aggregations => {
-            fetchTypeForTopEntitiesAggregation(aggregations, searchParameters);
-          }
-        );
+
+      if (resetAggregations) {
```
maniax89, line 297: i think the proper fix is probably to just move any calls to `handleSearch` with `resetAggregations = true` should just call `fetchAggregations` instead

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
kabeaty: I think let's not include this change for now, as it seems to cause a flicker for the category facets as multiselect before they have types.

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
kabeaty: Should we add a comment here?

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -287,31 +289,24 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
     ): Promise<void> => {
-      let aggregationsFetched = false;
       const searchParameters = deprecateReturnFields(
         backwardsCompatibleQueryParams
       ) as DiscoveryV2.QueryParams;
       setSearchParameters(searchParameters);
-      // don't use the search response if filter is set, just do another search
+
+      // TODO proper fix is to have any caller of handleSearch that needs aggregations
+      // to call handleFetchAggregations instead
```
jhpedemonte, line 298: Is the reason to not do this now to minimize the number of changes in this patch?

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -287,31 +289,24 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
-            ...aggregationQueryDefaults,
-            filter: ''
-          },
-          async aggregations => {
-            fetchTypeForTopEntitiesAggregation(aggregations, searchParameters);
-          }
-        );
+        const aggregationSearchParams = {
+          ...searchParameters,
+          ...aggregationQueryDefaults
```
jhpedemonte, line 302: Did we lose the utility function `resetAggregationParamsToDefaults`?

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
maniax89: yes

packages/discovery-react-components/src/components/SearchFacets/SearchFacets.tsx
maniax89: i added this back in with the non-storing of initial aggregations fetch - let me know if this works better now?

PR #182, opened 2021-02-12
feat: add error to dataapi

kabeaty:
This adds the `error` encountered to the state in `useDataApi`. This is so that it can be used in Discovery tooling.

Check that the tests look good. Can link with tooling and run tests added with the related changes there too.

no

### DIFF
--- a/packages/discovery-react-components/src/utils/__tests__/useDataApi.test.tsx
+++ b/packages/discovery-react-components/src/utils/__tests__/useDataApi.test.tsx
@@ -58,7 +58,11 @@ class AggregationResultSearchClient extends BaseSearchClient {
 
 class ErrorSearchClient extends BaseSearchClient {
   public async query(): Promise<any> {
-    return Promise.reject();
+    return Promise.reject(
+      new Error(
+        '"status_code":429,"errors":[{"code":"query_rate_limit","message":"Your query cannot be run because there are too many concurrent requests. Reduce the number of concurrent queries and try again.","more_info":"https://cloud.ibm.com/docs/discovery-data"}],"trace":"TOOLING-9947g26q0x1x9"}'
+      )
+    );
   }
 
   public async listFields(): Promise<any> {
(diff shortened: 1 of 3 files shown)


packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -172,6 +174,7 @@ interface ReducerState {
   data: any;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   parameters: any;
+  error: string;
```
maniax89, line 177: i think we are going to want to handle the various cases from the `formatError` function in the SDK - https://github.com/IBM/node-sdk-core/blob/main/lib/request-wrapper.ts#L260-L323

axios has a lot of weird ways it produces errors, I believe that this function tries to normalize them. The actual error object seems to look something like
```typescript
interface SdkError {
  message: string;
  statusText: string;
  body?: Record<string, any>;
  headers?: Record<string, any>;
};
```
maybe we store the entire error object instead and rely on this type to get the structured data out of the response rather than trying to parse a string?

packages/discovery-react-components/src/utils/__tests__/useDataApi.test.tsx
```diff
@@ -58,7 +58,11 @@ class AggregationResultSearchClient extends BaseSearchClient {
 
 class ErrorSearchClient extends BaseSearchClient {
   public async query(): Promise<any> {
-    return Promise.reject();
+    return Promise.reject(
+      new Error(
+        '"status_code":400,"errors":[{"code":"query_daily_limit","message":"You have exceeded the number of daily queries allowed for your plan. You can resume requests after the daily reset.","more_info":"https://cloud.ibm.com/docs/discovery-data"}],"trace":"TOOLING-9947g26q0x1x9"}'
```
maniax89, line 63: is this how the error is formatted? I think it might look something more like
```typescript
const error = new Error();
error.message = 'You have exceeded the number of daily queries allowed for your plan. You can resume requests after the daily reset.'
error.body = {status_code: 400, errors: [{code:'query_daily_limit',message: 'You have exceeded the number of daily queries allowed for your plan. You can resume requests after the daily reset.'}]};
```

packages/discovery-react-components/src/utils/useDataApi.ts
kabeaty: This makes sense to me, thank you. Updated!

maniax89 (approved):

packages/discovery-react-components/src/utils/useDataApi.ts
```diff
@@ -27,7 +27,8 @@ const dataFetchReducer = (state: any, action: any) => {
       return {
         ...state,
         isLoading: false,
-        isError: true
+        isError: true,
+        error: action.error || null
```
maniax89, line 31: i think for `FETCH_SUCCESS` and `FETCH_REQUEST` we might want to reset `error` to `null` otherwise it could linger around from previous requests

kabeaty: good point, done!

PR #191, opened 2021-05-24
docs: example for creating a custom component

jhpedemonte:

Add an example for creating a custom component.

Update example app as specified in the docs.

### DIFF
--- a/docs/components/custom_components.md
+++ b/docs/components/custom_components.md
@@ -1,0 +1,115 @@
+---
+layout: default
+title: Custom Components
+parent: Components
+permalink: /docs/Custom_Components
+---
+
+# Implementing a Custom Component
+
+The components we provide here may not fit your specific needs, and you'll need to write a custom component. Below is an example of how this could be done using React.
+
+## Scenario
+
+Let's say you want to use the Discovery's [Answer finding](https://medium.com/ibm-data-ai/finding-concise-answers-to-questions-in-enterprise-documents-53a865898dbd). Instead of returning _documents_, we want to get _answers_ from within those documents.
+
+## Implement a custom Answers component
+
+To make things easier, we're going to reuse the `discovery-search-app` example code -- specifically, within `src/App.js`. We'll keep using the search field, but create a custom component to show the top 3 answers.
+
+First, we need to update the query that is sent out when a user presses `Enter` in the search field. The value is set as the `natural_language_query` field of the request body. But we also need to enable _answers_, as described in the article. To do so, we can set the `overrideQueryParameters` property for `<DiscoverySearch>`:
+
+```diff
+   ) : isError ? (
+     <div>Unable to load Discovery projects. Please check your console for more details.</div>
+   ) : (
+-    <DiscoverySearch searchClient={searchClient} projectId={projectId}>
++    <DiscoverySearch
++      searchClient={searchClient}
++      projectId={projectId}
++      overrideQueryParameters={{
++        passages: {
++          enabled: true,
++          max_per_document: 3,
++          characters: 850,
++          fields: ['title', 'text'],
++          find_answers: true,
++          max_answers_per_document: 1
++        },
++        _return: ['document_passages', 'extracted_metadata']
++      }}
++    >
+       <AppView />
+     </DiscoverySearch>
+   );
+```
+
+The overall details are explained in the article, but I did want to point out a couple things. First, `fields` should be set to reflect the data in your project/collection. This property tells Discovery in which document fields to look for answers. Second, we're setting `_return` to a set of top-level fields in the document results. Only the fields listed will be returned, cutting down on the response size (especially since we're not listing `text` and `html`, which can be quite large).
+
+(Note: Besides `overrideQueryParameters`, you can also update the query parameters by calling the `setSearchParameters` function on the `SearchApi` context.)
+
+Now let's build our custom component. The Discovery Components use [React Context](https://reactjs.org/docs/context.html) to pass data and give access to their API. The results of a search are provided in the `searchResponseStore` property of `SearchContext`. From within that, we'll find the most relevant `answers`, based on their `confidence` score.
+
+Copy the following code into `src/App.js`:
+
+```javascript
+// Show top 3 answers from within search results
+function Answers() {
+  // Use the context provided by <DiscoverySearch> to get search results
+  const {
+    searchResponseStore: { data: searchResponse }
+  } = useContext(SearchContext);
+
+  if (!searchResponse) {
+    return null;
+  }
+
+  const top3Answers = searchResponse.results
+    // extract document passages from documents in the results...
+    .reduce((acc, doc) => acc.concat(doc.document_passages), [])
+    // extract answers from the passages...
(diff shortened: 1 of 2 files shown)


docs/components/custom_components.md
```diff
@@ -0,0 +1,115 @@
...
+permalink: /docs/Custom_Components
+---
+
+# Implementing a Custom Component
+
+The components we provide here may not fit your specific needs, and you'll need to write a custom component. Below is an example of how this could be done using React.
+
+## Scenario
+
+Let's say you want to use the Discovery's [Answer finding](https://medium.com/ibm-data-ai/finding-concise-answers-to-questions-in-enterprise-documents-53a865898dbd). Instead of returning _documents_, we want to get _answers_ from within those documents.
```
ColbyJohnIBM, line 14:
```suggestion
Let's say we want to use Watson Discovery's [Answer finding capability](https://medium.com/ibm-data-ai/finding-concise-answers-to-questions-in-enterprise-documents-53a865898dbd). Instead of returning _documents_, we want to get _answers_ from within those documents.
```

maclean-stu: This is a great idea..  Could you also consider writing the approach that customers could take to modify the existing search behaviour in terms of entities down the left had side to be a selection that results in an AND condition rather than an OR as is the current state. Consider a customer use case where they have a trained WKS model deployed to understand AIRCRAFT and SERIES entities and they want to bae able to drive search results based on a drop down list of AIRCRAFT = "one or ore selections"  AND SERIES = "one or more selections" .

michelle-miller: i do think you should stick with "you". (sorry to be contradictory, john) ibm style says, "Use the second-person pronoun _you_ as often as possible."
and we take extra care around references to product names. so rather than making Discovery possessive, and even though the link goes to info about the feature it might not hurt to add a short description. i recommend: "Let's say you want to use the Watson Discovery [answer finding] capability to highlight exact answers in query results"

docs/components/custom_components.md
```diff
@@ -0,0 +1,115 @@
+---
+layout: default
+title: Custom Components
+parent: Components
+permalink: /docs/Custom_Components
+---
+
+# Implementing a Custom Component
+
+The components we provide here may not fit your specific needs, and you'll need to write a custom component. Below is an example of how this could be done using React.
```
michelle-miller, line 10: here's a suggestion to avoid "we"
The components that are provided here might not fit your specific needs, and you'll need to write a custom component. 

and we try to avoid sensory words that assume sighted users. so, i try to replace terms like above and below whenever i see them.
The following example show you how to write a custom component by using React.

docs/components/custom_components.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Scenario
+
+Let's say you want to use the Discovery's [Answer finding](https://medium.com/ibm-data-ai/finding-concise-answers-to-questions-in-enterprise-documents-53a865898dbd). Instead of returning _documents_, we want to get _answers_ from within those documents.
+
+## Implement a custom Answers component
+
+To make things easier, we're going to reuse the `discovery-search-app` example code -- specifically, within `src/App.js`. We'll keep using the search field, but create a custom component to show the top 3 answers.
```
michelle-miller, line 18: might just be me, but i wasn't sure if we meant reuse the code from src/App.js or reuse from somewhere else and add it to src/App.js. so, i would suggest:
To make things easier, we're going to reuse the 'discovery-search-app' example code. We will copy it into the src/App.js file, and then make some changes to it.

docs/components/custom_components.md
```diff
@@ -0,0 +1,115 @@
...
++        },
++        _return: ['document_passages', 'extracted_metadata']
++      }}
++    >
+       <AppView />
+     </DiscoverySearch>
+   );
+```
+
+The overall details are explained in the article, but I did want to point out a couple things. First, `fields` should be set to reflect the data in your project/collection. This property tells Discovery in which document fields to look for answers. Second, we're setting `_return` to a set of top-level fields in the document results. Only the fields listed will be returned, cutting down on the response size (especially since we're not listing `text` and `html`, which can be quite large).
```
michelle-miller, line 47: maybe to avoid using I, we could word it as:
, but here are a few things to keep in mind.

docs/components/custom_components.md
```diff
@@ -0,0 +1,115 @@
...
+        {top3Answers.map(answer => (
+          <li>{answer.answer_text}</li>
+        ))}
+      </ol>
+    </div>
+  );
+}
+```
+
+(Note: no styling provided for this component.)
```
michelle-miller, line 90: should we say anything about what file they can use to add style info for this? 
Note: No style is defined for the Answers class (is it a class?), but you can add a style definition for it in the x file.

docs/components/custom_components.md
```diff
@@ -0,0 +1,115 @@
...
+-              {!isError ? <SearchResults /> : <p>An error occurred during search.</p>}
++              {!isError ? <Answers /> : <p>An error occurred during search.</p>}
+             </div>
+           </div>
+         </div>
+```
+
+## Write everything from scratch
+
+The example above uses a custom component alongside several existing Discovery Components. But you may wish to create everything from scratch. There's two ways to approach this.
```
michelle-miller, line 107: to avoid "above" and "alongside" which are sensory words (assume sighted users) i'd recommend: The previous example uses a custom component along with several ...
and slight reword suggestion for the next sentence:
But you might want to create everything from scratch. There are two ways to approach adding your own code.

docs/components/custom_components.md
jhpedemonte: I meant to just change `src/App.js` in place, not copy.

docs/components/custom_components.md
jhpedemonte: The way I see it, this is just a quick tutorial and I didn't want to get bogged down in unneeded details (like styling). Perhaps I should just remove this note entirely.

jhpedemonte:
@michelle-miller Updated based on your suggestions.

docs/components/custom_components.md
michelle-miller: just to prevent others from misunderstanding like i did, maybe:
To make things easier, we're going to edit the existing code for the 'discovery-search-app' in the src/App.js file.

michelle-miller (approved):

docs/components/custom_components.md
```diff
@@ -0,0 +1,113 @@
...
++          find_answers: true,
++          max_answers_per_document: 1
++        },
++        _return: ['document_passages', 'extracted_metadata']
++      }}
++    >
+       <AppView />
+     </DiscoverySearch>
+   );
+```
```
noah-eigenfeld, line 45: Real nice touch making this a diff instead of just a block of code

noah-eigenfeld (approved):
Instructions are concise and easy to understand 👏🏻

docs/components/custom_components.md
ColbyJohnIBM: I'm fine with "you"-- I was just wanting the first and second sentence to be consistent. :+1:

PR #195, opened 2021-07-01
fix: use "unstable" Carbon Pagination to avoid performance hang

jhpedemonte:

Carbon's `Pagination` component does not work well if there are a large number of items, resulting in app hanging or crashing (see https://github.com/carbon-design-system/carbon/issues/5546).

This PR changes us to use the "unstable" `Pagination` component from Carbon, which will get promoted to the official one in Carbon v11. By doing so, this no longer renders a select control to pick the current page (which is of dubious usability for search results), and avoids the performance issue.

New component:
Screen Shot 2021-07-01 at 3 48 39 PM

Easiest way to see the underlying issue is to load https://react.carbondesignsystem.com/?path=/story/components-pagination--pagination and enter a large number like `400000` for `totalItems` (in "Knobs" tab at bottom of page). Browser page will hang.

To see fix, run our Storybook with this PR and load `ResultsPagination`. Notice that there is no longer a select control for the current page (on right-hand side of pagination bar).

### DIFF
--- a/packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
+++ b/packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
@@ -83,7 +83,7 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
     }
   }, [currentPage, searchParameters.count, searchParameters.offset]);
 
-  const matchingResults = (searchResponse && searchResponse.matching_results) || 0;
+  const matchingResults = (searchResponse && searchResponse.matching_results) || undefined;
   const actualPageSize = searchParameters.count || 10;
   // the default behavior of Carbon is to discard pageSize if it is not included in pageSizes,
   // we instead choose to make it so that pageSize is appended to pageSizes if it is not already included.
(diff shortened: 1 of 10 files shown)


packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -50,7 +50,7 @@ interface ResultsPaginationEvent {
 const ResultsPagination: FC<ResultsPaginationProps> = ({
   page = 1,
   pageSizes = [10, 20, 30, 40, 50],
-  pageSize,
+  pageSize = 10,
```
jhpedemonte, line 53: Would default to 5 if not set (for some reason), even though that wasn't one of the `pageSizes` specified. So I changed it to default to 10.

noah-eigenfeld (approved):
Interesting workaround. Didn't test locally, but code changes look good.

noah-eigenfeld:
Local testing result: I may need to do something to update styling, but the pagination component looks like this for me on Chrome
Screen Shot 2021-07-22 at 10 34 23 AM

noah-eigenfeld:
Update: after rebuilding components, styling looks fine for me. Ignore my comment.
Screen Shot 2021-07-22 at 10 46 04 AM

noah-eigenfeld:
After testing locally, still approved 👍🏻

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -146,6 +148,48 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
   return null;
 };
 
+// XXX Slight hack. unstabled_Pagination doesn't currently emit an `onChange`
```
ColbyJohnIBM, line 151: Don't love that this hack is necessary, but nice workaround.

ColbyJohnIBM (approved):
:100:

tusharcoepit:
@jhpedemonte - Is `onChange` triggering for you with experimental pagination component ? For us it is not and we are seeing an issue with page change -https://github.com/carbon-design-system/ibm-security/issues/1051

Any pointers / references you can share to analyse if we are missing something. Thanks!

jhpedemonte:
@tusharcoepit No, `onChange` is not working as expected with this new component. But I did find a [workaround](https://github.com/watson-developer-cloud/discovery-components/pull/195/files#diff-1e32489748d90317ba0697bfd33de5951a89aff2e9874c6722ab24f942e8195aR151).

PR #208, opened 2021-09-02
fix: better tree shaking with @carbon/icons-react and carbon-react-co…

### LINKED ISSUE #202: discovery-react-components includes all of the @carbon/icons-react package rather than tree-shaking
When importing the discovery-react-components into my react app, the package results in the entirety of the @carbon/icons-react package being included in my webpack bundle (5.76 MB) rather than allowing it to be tree shaken out.
I believe its because the discovery-react-components import the icons from the lib path rather than the suggested way like so on line 14 in the SearchInput component.
```
import Search16 from '@carbon/icons-react/lib/search/16';
```
rather than
```
import { Search16 } from '@carbon/icons-react';
```

See this issue for more details on importing carbon icons from the lib path. https://github.com/carbon-design-system/carbon/issues/4847

psulkava:

discovery-react-components does not properly import icons from @carbon/icons-react and carbon-react-components ListBox to allow for tree shaking. This changes these imports to use the /es/ path which allows them to be tree shaken.

fixes #202 

Import discovery-react-components into a new create-react-app app and use webpack-bundle-analyzer to see the vendors.js bundle size.

No.

I don't believe so.


packages/discovery-react-components/src/components/CIDocument/components/NavigationToolbar/NavigationToolbar.tsx
```diff
@@ -2,8 +2,8 @@ import React, { FC } from 'react';
 import cx from 'classnames';
 import { settings } from 'carbon-components';
 import { Button } from 'carbon-components-react';
-import ChevronLeft16 from '@carbon/icons-react/lib/chevron--left/16';
-import ChevronRight16 from '@carbon/icons-react/lib/chevron--right/16';
+import ChevronLeft16 from '@carbon/icons-react/es/chevron--left/16';
+import ChevronRight16 from '@carbon/icons-react/es/chevron--right/16';
```
jhpedemonte, line 6: If we're making these changes anyway, I would prefer to go with the cleaner looking code:
```suggestion
import { ChevronRight16 } from '@carbon/icons-react';
```

If you don't have time to update this, I can handle it tomorrow.

packages/discovery-react-components/src/typings.d.ts
```diff
@@ -20,7 +20,7 @@ declare module '@carbon/icons-react/*';
 declare module '*.md';
 declare module 'marked';
 declare module 'uuid';
-declare module 'carbon-components-react/lib/components/ListBox';
+declare module 'carbon-components-react/es/components/ListBox';
```
jhpedemonte, line 23: I think the reason this is here is due to including this specific component in `rollup.config.js`. I don't remember why that was added there. That ref should be changed to `/es/` as well.

jhpedemonte (commented):
Looks good. I verified by building the example app, which had one chunk decrease from 875K to 508K.

packages/discovery-react-components/src/typings.d.ts
jhpedemonte: Oh, I see: we're using `ListBox` but it's not exported at the "top level" by carbon components.

packages/discovery-react-components/src/components/CIDocument/components/NavigationToolbar/NavigationToolbar.tsx
psulkava: I tried using that cleaner format but for some reason I couldn't get it to build that way. It only worked for me when using the full /es/ path. If you want to take a look and see if you can figure out why you probably understand it better than I.

packages/discovery-react-components/src/typings.d.ts
psulkava: Right. I'm not very familiar with Typescript but assumed this needed to be updated as well when I updated the path for the ListBox.

jhpedemonte:
Closed in favor of #209

PR #229, opened 2021-10-25
fix: PdfViewer scale parameter issue

fukudasjp:

When you specify `scale` props to `PdfViewer` componenet, the PDF image on the screen scales by the square of the `scale` value.

For example, when you specify `1.5` to the value, the width of the image gets `1.5 * 1.5 = 2.25` times of the original. That's because the `canvas` size is determined with `scale` and the `scale` is again used to scale the `canvas` by `scale`.

Use an existing storybook.

no


fukudasjp:
Moved to #232

PR #230, opened 2021-10-27
feat: use device pixel ratio to render PDF

fukudasjp:

Currently, PDFViewer doesn't consider device pixcel ratio and rendered PDF pages are not clear on Mac. This PR is to consider device pixcel ratio to render PDF image.

Before
image
After
image

Verify the rendered PDF on Storybook. _Use device resolution_ options is added to the storybook for the previous behavior.


fukudasjp:
Moved to #233 to use local branch

PR #231, opened 2021-10-28
fix: correct query filter string generation

noah-eigenfeld:
Implements: https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/9339

Fixes the default logic assembling query filter string anytime that facets are applied to a search.

~This change should add parens ( ) around the values for any fields in the filter param. That way, if there are multiple values for one facet, they are correctly bundled together. Hopefully, it should look something like `facet-name:("value1"|"value2"|"value3"),facet-2-name:("value4"), . . .`.~

The parens I tried to add were getting stripped out by a followup step that gets rid of special characters. Instead, I opted to stack the facet-name/value pairs. They should now look like this:
```
filter: facet-name:"value1"|facet-name:"value2"|facet-name:"value3",facet-2-name:"value4", . . .
```

1. Start the example app
2. Perform a search
3. Filter results with the facets panel
4. See that the query request that gets sent out when you change facets has the correctly formatted `filter` string
N/A
No

### DIFF
--- a/packages/discovery-react-components/src/components/SearchFacets/utils/__tests__/searchFilterTransform.test.ts
+++ b/packages/discovery-react-components/src/components/SearchFacets/utils/__tests__/searchFilterTransform.test.ts
@@ -113,11 +113,6 @@ describe('Filter string to QueryTermAggregation array', () => {
   });
 
   test('it properly handles aggregations with reserved characters', () => {
-    //TODO: failing
-    console.log(
-      'Received filters value: ',
-      SearchFilterTransform.fromString(weirdFilters).filterFields[1]
-    );
     expect(SearchFilterTransform.fromString(weirdFilters).filterFields).toEqual([
       {
         type: 'term',
--- a/packages/discovery-react-components/src/components/SearchFacets/utils/__tests__/searchFilterTransform.test.ts
+++ b/packages/discovery-react-components/src/components/SearchFacets/utils/__tests__/searchFilterTransform.test.ts
@@ -77,10 +77,10 @@ const weirdAggs = [
   }
 ];
 const weirdFilters =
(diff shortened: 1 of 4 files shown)


packages/discovery-react-components/src/components/SearchFacets/utils/__tests__/searchFilterTransform.test.ts
```diff
@@ -112,14 +113,20 @@ describe('Filter string to QueryTermAggregation array', () => {
   });
 
   test('it properly handles aggregations with reserved characters', () => {
+    //TODO: failing
+    console.log(
```
noah-eigenfeld, line 117: Gotta remove these

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
```diff
@@ -79,7 +83,7 @@ export class SearchFilterTransform {
       const results = get(facet, 'results', []);
       const keys = this.quoteSelectedFacets(results, 'key');
       if (keys.length) {
-        filterStrings.push(`${escapeFieldName(field)}:${keys.join('|')}`);
+        filterStrings.push(keys.map(key => `${escapeFieldName(field)}:${key}`).join('|'));
```
noah-eigenfeld, line 86: This is the main functional change for this PR

broulaye (approved):
Checked it locally, looks good to me 👍

jhpedemonte (approved):
Tested using example app (tip: had to delete `/examples/discovery-search-app/node_modules` and re-run `yarn` in order to pick up latest changes). In a project with the same file under 2 different names and a facet on `extracted_metadata.filename`, selecting both files in the facet on an empty search showed both results (previously, would only show the first file).

PR #234, opened 2021-11-08
chore: yarn v2

jhpedemonte:

Update to use yarn v2.

Run `yarn` to install dependencies. Then try building components and running example app.

---

### To do
- [x] `npm publish`


package.json
```diff
@@ -106,16 +106,6 @@
...
   },
-  "resolutions": {
-    "react-scripts/**/url-parse": "^1.5.1",
-    "@storybook/addon-docs/**/pug": "^3.0.2",
-    "webpack/**/elliptic": "^6.5.4",
-    "**/hosted-git-info": "^3.0.8",
-    "**/lodash": "^4.17.21",
-    "**/react-dev-utils": "^11.0.4",
-    "**/react-syntax-highlighter/**/prismjs": "^1.23.0",
-    "**/ssri": "^8.0.1"
```
jhpedemonte, line 117: Ran `yarn why` on each of these pkgs and they were all _at least_ the version listed here.

package.json
```diff
@@ -6,8 +6,12 @@
...
   ],
   "scripts": {
-    "lint": "wsrun -m -e -c lint",
-    "audit": "wsrun -m -e -c audit"
+    "lint": "yarn workspaces foreach -v --parallel run lint",
+    "audit": "yarn workspaces foreach -v --parallel run audit",
+    "build:pkgs": "yarn workspaces foreach -v --parallel --from '@ibm-watson/*' run build",
+    "build:apps": "yarn workspaces foreach -v --parallel run build:app",
+    "test": "yarn workspaces foreach -v --parallel run test",
+    "g:eslint": "cd $INIT_CWD && eslint"
```
jhpedemonte, line 14: This is a "global" script, which is callable from anywhere in the workspace (currently used in `examples/discovery-search-app/package.json`). This allows us to install `eslint` pkg at the top level, then use in multiple places. Otherwise, yarn v2 would require that each workspace include `eslint` as a dependency -- this would make maintenance cumbersome, as we'd have to update the `eslint` version in multiple places.

See https://yarnpkg.com/getting-started/qa#how-to-share-scripts-between-workspaces.

.github/workflows/ci.yaml
```diff
@@ -25,15 +25,15 @@ jobs:
         with:
           node-version: 12
       - name: Install
-        run: yarn install --frozen-lockfile
+        run: yarn install --immutable
       - name: Lint
-        run: npx lerna run lint --stream
+        run: yarn run lint
```
jhpedemonte, line 30: 1. `lerna` doesn't work with the `workspace:` (see below)
2. yarn v2 has a plugin which allows running a command in all/most workspaces (`yarn workspaces foreach`). These yarn scripts here make use of that.

examples/discovery-search-app/package.json
```diff
@@ -23,8 +23,8 @@
   },
   "dependencies": {
     "@carbon/icons": "^10.5.0",
-    "@ibm-watson/discovery-react-components": "file:../../packages/discovery-react-components",
-    "@ibm-watson/discovery-styles": "file:../../packages/discovery-styles",
+    "@ibm-watson/discovery-react-components": "workspace:*",
+    "@ibm-watson/discovery-styles": "workspace:*",
```
jhpedemonte, line 27: yarn v2 supports a new protocol `workspace:`, which finds a package in the local workspaces, rather than pulling from remote repo. Using `*` tells yarn to find a package by the given name.

See https://yarnpkg.com/features/protocols#workspace

.github/workflows/ci.yaml
jhpedemonte: Decided to get rid of `workspace:` protocol, but kept these changes in order to lessen our dependence on `lerna` (only used for `publish`).

examples/discovery-search-app/package.json
```diff
@@ -23,8 +22,8 @@
   },
   "dependencies": {
     "@carbon/icons": "^10.5.0",
-    "@ibm-watson/discovery-react-components": "file:../../packages/discovery-react-components",
-    "@ibm-watson/discovery-styles": "file:../../packages/discovery-styles",
+    "@ibm-watson/discovery-react-components": "^1",
+    "@ibm-watson/discovery-styles": "^1",
```
jhpedemonte, line 26: From the [yarn docs](https://yarnpkg.com/features/workspaces):

> If the package manager was to resolve a range that a workspace could satisfy, it will prefer the workspace resolution over the remote resolution if possible.

I tried using `workspace:` protocol first, but that broke `lerna`. This approach here should mean that the example app _should_ pull in these packages from the local workspaces, rather than pulling from remote repo.

examples/discovery-search-app/package.json
```diff
@@ -23,8 +22,8 @@
   },
   "dependencies": {
     "@carbon/icons": "^10.5.0",
-    "@ibm-watson/discovery-react-components": "file:../../packages/discovery-react-components",
-    "@ibm-watson/discovery-styles": "file:../../packages/discovery-styles",
+    "@ibm-watson/discovery-react-components": "^1.5.0-beta.0",
+    "@ibm-watson/discovery-styles": "^1.4.1-beta.1",
```
jhpedemonte, line 26: Can't use `workspace:` or `portal:`, since those aren't supported by lerna. `file:` was also giving me issues, resulting in a different `yarn.lock` on Travis vs locally. In the end, just added the current versions -- yarn will actually map these on to the local workspace folder, since the versions are compatible.

package.json
noah-eigenfeld: Neat!

packages/discovery-react-components/package.json
```diff
@@ -40,26 +59,12 @@
     "react-dom": ">= 16.8.0 < 17"
   },
   "engines": {
-    "node": ">=8",
-    "npm": ">=5"
+    "node": ">=12"
```
noah-eigenfeld, line 62: 👍🏻

noah-eigenfeld (approved):

PR #237, opened 2021-11-09
feat: add text layer support to PDF viewer

fukudasjp:
Add `showTextLayer` option to render text layer on `PDFViewer` as it's necessary for implementing highlighting on PDF. It would be useful for text selection and in-document search feature for future. Users can disable text layer rendering when they don't need it (for such as rendering mini-map images).

For implementation, this PR does the following things, which are tied to commits:
- Add `@types/pdfjs-dist` for better code
- Control PDF rendering tasks properly in `PDFViewer` (properly handles cancelled rendering)
- Add `PDFViewerTextLayer` component
   - Update the storybook with`Show text layer` option

- Use storybook to test the feature

- No

no


packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -33,21 +54,30 @@ interface Props {
...
+  setTextLayerInfo,
+  children
 }) => {
   const canvasRef = useRef<HTMLCanvasElement>(null);
 
   // In order to prevent unnecessary re-loading, loaded file and page are stored in state
-  const [loadedFile, setLoadedFile] = useState<any>(null);
-  const [loadedPage, setLoadedPage] = useState<any>(null);
+  const [loadedFile, setLoadedFile] = useState<PDFDocumentProxy | null>(null);
+  const [loadedPage, setLoadedPage] = useState<PDFPageProxy | null>(null);
```
jhpedemonte, line 80: 🎉

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -88,33 +118,74 @@ const PdfViewer: SFC<Props> = ({
...
   }, [loadedFile, page]);
 
-  const [viewport, canvasInfo] = useMemo(() => {
-    const viewport = loadedPage?.getViewport({ scale });
-    const canvasInfo = viewport ? getCanvasInfo(viewport) : undefined;
-    return [viewport, canvasInfo];
-  }, [loadedPage, scale]);
+  const currentPage = useMemo(() => {
+    const isPageValid = !!loadedPage && loadedPage.pageNumber === page;
+    if (isPageValid) {
```
jhpedemonte, line 123: What's the reason for checking if the page is "valid"? We control `loadedPage` and it will either be `null` or a legit PDF page. So seems like all we need to check is if it's non-null, right?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -88,33 +118,74 @@ const PdfViewer: SFC<Props> = ({
...
+      const viewport = loadedPage?.getViewport({ scale });
+      const canvasInfo = viewport ? getCanvasInfo(viewport) : undefined;
+      return { loadedPage, viewport, canvasInfo };
```
jhpedemonte, line 126: This will make the code confusing, as in the following section: there's the main `loadedPage`, and then there's the one from within `currentPage`.

Actually, there isn't a reason to store `loadedPage` within `currentPage`, since it's already available in this context. Is there a reason for going away from the previous code, which added `viewport` and `canvasInfo` at the main closure level?

If there is a reason for that, then I suggest we instead name this `loadedPageInfo` and set it to `{ viewport, canvasInfo }`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -88,33 +118,74 @@ const PdfViewer: SFC<Props> = ({
     };
   }, [loadedFile, page]);
 
-  const [viewport, canvasInfo] = useMemo(() => {
-    const viewport = loadedPage?.getViewport({ scale });
-    const canvasInfo = viewport ? getCanvasInfo(viewport) : undefined;
-    return [viewport, canvasInfo];
-  }, [loadedPage, scale]);
+  const currentPage = useMemo(() => {
```
jhpedemonte, line 121: Let's add comments to each of these `useMemo`/`useEffect` sections, to document the _intent_ of the code. This will make it easier to reason about how all of these fit together.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -88,33 +118,74 @@ const PdfViewer: SFC<Props> = ({
...
+
+    const { loadedPage, viewport, canvasInfo } = currentPage || {};
+    if (loadedPage && !(loadedPage as any).then && viewport && canvasInfo) {
+      const render = async () => {
+        try {
+          task = _renderPage(loadedPage, canvasRef.current!, viewport, canvasInfo);
+          await task?.promise;
+        } catch (e) {
+          if (e instanceof RenderingCancelledException) {
+            // ignore
```
jhpedemonte, line 143: Why do we ignore this? Please add more to this comment.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -88,33 +118,74 @@ const PdfViewer: SFC<Props> = ({
...
+            setLoading(false);
+          }
+        }
+      };
+      render();
     }
-  }, [loadedPage, viewport, canvasInfo, setLoading]);
+    return () => {
+      didCancel = true;
+      task?.cancel();
```
jhpedemonte, line 157: Nice!

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -88,33 +118,74 @@ const PdfViewer: SFC<Props> = ({
...
+  }, [loadedPage, currentPage, setLoading]);
 
   useEffect(() => {
     if (setHideToolbarControls) {
       setHideToolbarControls(false);
     }
   }, [setHideToolbarControls]);
 
+  const classNameBase = `${settings.prefix}--document-preview-pdf-viewer`;
+  const { loadedPage: currentLoadedPage, canvasInfo } = currentPage || {};
```
jhpedemonte, line 168: Again, this seems unnecessary to rename `loadedPage` here -- just use the existing var.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+    page: number;
+    scale: number;
+    textContent: TextContent;
+    viewport: PDFPageViewport;
+  } | null>(null);
+
+  useEffect(() => {
+    async function loadTextInfo() {
+      const isPageReady = !!loadedPage && loadedPage.pageNumber === page;
+      if (isPageReady) {
```
jhpedemonte, line 77: Same comment as in previous file.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+  page = 1,
+  scale = 1,
+  setTextLayerInfo: setTextLayerInfoCallback = () => {}
+}) => {
+  const textLayerRef = useRef<HTMLDivElement>(null);
+  const textLayerDiv = textLayerRef.current;
+
+  const [textRenderInfo, setTextRenderInfo] = useState<{
+    page: number;
+    scale: number;
```
jhpedemonte, line 69: These are incoming props. There's no reason to store them in state.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+          }
+        }
+      }
+      setTextLayerInfo(textLayerInfo);
+    }
+    loadTextLayer();
+
+    return () => {
+      textLayerBuilder?.cancel();
+      // should we set text items??
```
jhpedemonte, line 149: Are you going to handle this comment in this PR?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+          textLayerInfo = {
+            textContent,
+            textDivs: textLayerBuilder.textDivs,
+            viewport,
+            page
+          };
+        } catch (e) {
+          if (e instanceof RenderingCancelledException) {
+            // ignore
```
jhpedemonte, line 136: Same comment as in previous file.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+        textLayerBuilder.setTextContent(textContent);
+
+        // render
+        textLayerDiv.innerHTML = '';
+        try {
+          const deferredRenderEnd = (() => {
+            let resolve: null | Function = null;
+            const promise = new Promise(resolvePromise => {
+              resolve = resolvePromise;
+            });
```
jhpedemonte, line 111: This is an odd way of doing this. Why not do this the standard way?

```js
const deferredRenderEndPromise = (() => {
  return new Promise(resolve => {
    const listener = () => {
      resolve();
      textLayerBuilder?.eventBus.off('textlayerrendered', listener);
    };
    textLayerBuilder.eventBus.on('textlayerrendered', listener);  
  });
})();
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+        const textContent = await loadedPage.getTextContent();
+        setTextRenderInfo({ textContent, viewport, page, scale });
+      }
+    }
+    loadTextInfo();
+  }, [loadedPage, page, scale]);
+
+  const textLayerBuilderRef = useRef<TextLayerBuilder | null>(null); // ref for debugging purpose
+  const [textLayerInfo, setTextLayerInfo] = useState<PdfTextLayerInfo | null>(null);
+  useEffect(() => {
```
jhpedemonte, line 88: This `useEffect` is quite large. I suggest refactoring this to a standalone function or even a hook (you can still have this within this same file; no reason to make a separate file for either approach).

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,207 @@
...
+      style={{
+        width: `${textRenderInfo?.viewport?.width ?? 0}px`,
+        height: `${textRenderInfo?.viewport?.height ?? 0}px`
+      }}
+    />
+  );
+};
+
+/**
+ * Adjust text span width
```
jhpedemonte, line 171:
```suggestion
 * Adjust text span width based on scale
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/typings.d.ts
```diff
@@ -1,2 +1,54 @@
-declare module 'pdfjs-dist';
 declare module 'pdfjs-dist/build/pdf.worker.min.js';
+
+//
+// Declare modules and their types that is referred from PDF text layer rendering.
+// Unused properties are commented out
+//
+declare module 'pdfjs-dist/lib/web/ui_utils' {
```
jhpedemonte, line 7: Are none of these included in the `@types` pkg you added?

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
```diff
@@ -1,3 +1,61 @@
...
+.#{$prefix}--document-preview-pdf-viewer--canvas {
+  transform-origin: left top 0px;
+}
+
+.#{$prefix}--document-preview-pdf-viewer--text {
   transform-origin: left top 0px;
+
+  //
+  // NOTE: import textLayer styles from ~pdfjs-dist/web/pdf_viewer.css
+  // @import "~pdfjs-dist/web/pdf_viewer" doesn't work for loading image
```
jhpedemonte, line 14: Can you explain this? I'd rather not do this at all, since it will make maintenance difficult (and we'll need to update PDF.js pkg soon).

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
```diff
@@ -1,3 +1,61 @@
...
+    white-space: pre;
+    cursor: text;
+    transform-origin: 0% 0%;
+  }
+
+  &.textLayer ::selection {
+    background: rgba(0, 0, 255, 1);
+  }
+
+  /* Avoids https://github.com/mozilla/pdf.js/issues/13840 in Chrome */
```
jhpedemonte, line 41: Please add more to this comment. In particular, a quick summary of the actual issue you're trying to work around. A reader shouldn't need to view that URL, unless they want to view more details. Something like:
```suggestion
  // Avoids ...some bug... in Chrome
  // @see https://github.com/mozilla/pdf.js/issues/13840
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -21,6 +32,16 @@ interface Props {
    */
   scale: number;
 
+  /**
+   * Render text layer
+   */
+  showTextLayer?: boolean;
```
jhpedemonte, line 38: Is there a reason for making this an option? Why not _always_ show the text layer when rendering a PDF?

jhpedemonte (commented):
Tested locally in Storybook and works great. But have many suggestions/questions for the code.

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
jhpedemonte: If you need help with this, let me know and I can dig in as well.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
fukudasjp: I think that we reuse this component to render thumbnail image or tile.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/typings.d.ts
fukudasjp: Right. The `@types/pdfjs-dist` only includes typings for _core_ classes.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
fukudasjp: The purpose is doing this is to make sure the page is loaded exactly once when user opens a document or changes a page. But I agree that having `loadedPage` and `currentPage` makes the code confusing. So, let me think about refactoring the logic.

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
fukudasjp: Thanks! I made this due to two reasons:
- the css file includes more styles other than ones for text layer. this is not critical
- when I tried `@import "~pdfjs-dist/web/pdf_viewer"`, `yarn run storybook` is blocked by errors like `Can't resolve './images/loading-icon.gif'`. The gif is in the `~pdfjs-dist/web/pdf_viewer/images`. I couldn't make it work. So it 's great help if you could take a look...

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
fukudasjp: I had to do this tweak to make sure all the variables are updated in synchronous. As we have to use async method to load PDF text content, there will be timing gap between the time `loadedPage` is updated and the time `textContent` is updated. It can triggers text rendering in the middle of updating states. 

Maybe we can overcome it by having one large async method, but I thought that it's worth splitting`getTextConten` from PDF and `renderTextContent`. I appreciate your suggestion.

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
fukudasjp: Did some more investigation
- When we import the styles as scss from another module, we don't have issue because node-sass (and webpack) resolve the dependency in the module
- When we try to import the compiled css `css/index.css`, the image link is broken in the css file and we get error

What we can do would be:
- copy the images to `css/images/` folder in the build process. But I think this package is only for css and 'm not sure if it's good idea to include images or not

fukudasjp: I found:
- Importing the SCSS from an external module, it works
- Importing the SCSS from the storybook, it results in error

So, I'm guessing it's due to webpack configuration.

jhpedemonte: Oh, I see. `pdf_viewer.css` doesn't have any "scoped" CSS -- it applies globally. Copying them into this parent SCSS rule takes care of that.

Also, `pdf_viewer.css` contains a lot more than just the lines you copied here.

However, I'm still concerned about this, since this makes maintenance complex -- whenever we upgrade `pdfjs-dist`, we have to remember that this code was copied and update it (if the original changed). Perhaps a better solution would be someway of copying it as a build step? I haven't fully thought through the best solution.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
jhpedemonte: I'm worried that this would take up a lot of resources, if we need to create many "thumbnails". We'll have to test this out and make sure we don't overwhelm the browser.

SDU Annotator has a server-side process to create the thumbnails. I don't believe we will be able to make use of that, but a similar approach may be needed here.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -1,11 +1,22 @@
...
+  PDFDocumentProxy,
+  PDFPageProxy,
+  PDFPageViewport,
+  PDFRenderTask
+} from 'pdfjs-dist';
 import PdfjsWorkerAsText from 'pdfjs-dist/build/pdf.worker.min.js';
 import { settings } from 'carbon-components';
+import PdfViewerTextLayer, { PdfTextLayerInfo } from './PdfViewerTextLayer';
+
+const { RenderingCancelledException } = PdfjsLib as any;
```
jhpedemonte, line 13: Can you add `RenderingCancelledException` as a type in the types file? I think you can just make it of type `Error`, right?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,215 @@
+import React, { FC, useEffect, useRef, useState } from 'react';
+import cx from 'classnames';
+import PDFJSLib, { PDFPageProxy, PDFPageViewport, TextContent, TextContentItem } from 'pdfjs-dist';
+import { EventBus } from 'pdfjs-dist/lib/web/ui_utils';
+import { TextLayerBuilder } from 'pdfjs-dist/lib/web/text_layer_builder';
+
+const { RenderingCancelledException } = PDFJSLib as any;
+
+interface Props {
```
jhpedemonte, line 9: Are these just a subset of the PdfViewer props? If so, should this just extend those? Something like:

```js
type Props = Pick<PdfViewerProps, 'loadedPage' | 'scale'> & { setTextLayerInfo?: ... };
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,215 @@
...
+  textContent: TextContent;
+
+  /** @see Props['scale'] */
+  scale: number;
+
+  /** @see PdfTextLayerInfo['viewport'] */
+  viewport: PDFPageViewport;
+
+  /** @see PdfTextLayerInfo['page'] */
+  page: number;
```
jhpedemonte, line 63: Same with these. If these are the same as from other types, then use `Pick<>`.

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
fukudasjp: OK. for the first step, I extracted the styles to a new file `_pdfjs_web_mixins.scss` and import the styles in `_document-preview-pdf-viewer.scss`.

I'll create a script to update the `_pdfjs_web_mixins.scss` later

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
fukudasjp: Only `scale` prop is so. I updated the type.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
fukudasjp: this type is removed after refactoring :)

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
fukudasjp: Added `yarn run prebuild` script to sync the `_pdfjs_web_mixins.scss` with the latest css from pdfjs-dist package.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
jhpedemonte: Did this change? I see same `Props` object as previously (and my comment above still stands).

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/useAsyncFunctionCall.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from 'react';
```
jhpedemonte, line 1: This hook isn't specific to `PdfViewer`, so it should be moved up to the top-level `utils` dir.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/useAsyncFunctionCall.ts
```diff
@@ -0,0 +1,46 @@
...
+ *  Take one parameter `setCancellable` to set _cancellable_ of the current async call.
+ * @returns the result of the async function
+ */
+function useAsyncFunctionCall<Func extends AsyncFunc<any>, ReturnType = AsyncFuncReturnType<Func>>(
+  asyncFunction: Func
+): ReturnType | undefined {
+  const [result, setResult] = useState<ReturnType | undefined>();
+
+  useEffect(() => {
+    let state: 'pending' | 'fulfilled' | 'rejected' = 'pending';
```
jhpedemonte, line 19: Nit: seems this function only needs to differentiate between `pending` or not, so this could be simplified to just a simple boolean.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/useAsyncFunctionCall.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from 'react';
+
+type AsyncFunc<ReturnType> = (signal: AbortSignal) => Promise<ReturnType>;
+type AsyncFuncReturnType<T> = T extends AsyncFunc<infer U> ? U : never;
+
+/**
+ * Call async function WRAPPED BY `useCallback` and return its result
```
jhpedemonte, line 7: So every instance of this _requires_ that the caller use `useCallback`. Why not just bring that into this hook? And call it `useAsyncCallback`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -33,88 +53,91 @@ interface Props {
...
-  }, [loadedFile, page]);
+  const loadedFile = useAsyncFunctionCall(
+    useCallback(async () => (file ? await _loadPdf(file) : null), [file])
+  );
+  const loadedPage = useAsyncFunctionCall(
+    useCallback(
+      async () => (loadedFile && page > 0 ? await _loadPage(loadedFile, page) : null),
+      [loadedFile, page]
+    )
+  );
```
jhpedemonte, line 85: Nice cleanup!

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,179 @@
...
+        if (textLayerDiv && loadedText) {
+          const { textContent, viewport, scale, page } = loadedText;
+
+          const builder = new TextLayerBuilder({
+            textLayerDiv,
+            viewport,
+            eventBus: new EventBus(),
+            pageIndex: page - 1
+          });
+          signal.addEventListener('abort', () => builder.cancel());
```
jhpedemonte, line 85: I'm really liking this AbortController usage pattern. We should consider bringing this in to main Tooling (as an import from this library).

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,179 @@
...
+  textLayerDiv: HTMLDivElement,
+  scale: number
+) {
+  builder.setTextContent(textContent);
+
+  // render
+  textLayerDiv.innerHTML = '';
+  const deferredRenderEndPromise = new Promise(resolve => {
+    const listener = () => {
+      resolve(undefined);
```
jhpedemonte, line 130: Why pass `undefined` instead of just calling `resolve()`?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -0,0 +1,179 @@
...
+async function _renderTextLayer(
+  builder: TextLayerBuilder,
+  textContent: TextContent,
+  textLayerDiv: HTMLDivElement,
+  scale: number
+) {
+  builder.setTextContent(textContent);
+
+  // render
+  textLayerDiv.innerHTML = '';
```
jhpedemonte, line 127: Is this to clear out the contents of `textLayerDiv`?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/typings.d.ts
```diff
@@ -1,2 +1,54 @@
...
+  export class EventBus {
+    on(eventName: string, listener: any): void;
+    off(eventName: string, listener: any): void;
+    dispatch(eventName: string, args?: any): void;
+  }
+  // export function getGlobalEventBus(): EventBus;
+}
+
+declare module 'pdfjs-dist/lib/web/text_layer_builder' {
```
jhpedemonte, line 16: Some (all?) of these types seem to be defined by `pdfjs-dist` itself: https://github.com/mozilla/pdfjs-dist/blob/master/types/web/text_layer_builder.d.ts

packages/discovery-react-components/package.json
```diff
@@ -43,6 +43,7 @@
     "react-virtualized": "9.21.1"
   },
   "devDependencies": {
+    "@types/pdfjs-dist": "^2.10.378",
```
jhpedemonte, line 46: It seems that this [package](https://www.npmjs.com/package/@types/pdfjs-dist) is now deprecated, showing:
> This is a stub types definition. pdfjs-dist provides its own type definitions, so you do not need this installed.

packages/discovery-styles/package.json
```diff
@@ -7,6 +7,7 @@
   "repository": "https://github.com/watson-developer-cloud/discovery-components",
   "main": "scss/index.scss",
   "scripts": {
+    "prebuild": "scripts/update-styles.sh",
```
jhpedemonte, line 10: Does `prebuild` script actually get called? Things changed with yarn v2 and some of these no longer get run automatically. See this [migration item](https://yarnpkg.com/getting-started/migration#explicitly-call-the-pre-and-post-scripts). I also don't see `prebuild` mentioned in the supported [lifecycle scripts](https://yarnpkg.com/advanced/lifecycle-scripts).

packages/discovery-styles/scripts/update-styles.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
```
jhpedemonte, line 2: Please add a comment block here explaining what this script does and how to invoke it (usage).

packages/discovery-styles/scripts/update-styles.sh
```diff
@@ -0,0 +1,20 @@
...
+  
+  sed -e "/BEGIN-QUOTE $key/q" $file > $tmp
+  cat >> $tmp
+  sed -ne "/END-QUOTE $key/,\$p" $file >> $tmp
+  cp $tmp $file;
+  rm $tmp;
+}
+
+cat $PDFJS_WEB_CSS | awk '/^\/\*/,/\*\//' | replace_quote $PDFJS_SCSS "COMMENT"
+cat $PDFJS_WEB_CSS | awk '/^\.textLayer/,/}/' | replace_quote $PDFJS_SCSS "TEXT-LAYER"
```
jhpedemonte, line 19: I'm a bit concerned about using string manipulation for these style script files. There's existing utilities (like `eslint`/`prettier`) which will correctly read in the files and make them available to manipulation in a safe way. Writing a plugin for such a utility would be better.

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
```diff
@@ -1,3 +1,14 @@
+@import './pdfjs_web_mixins';
+
 .#{$prefix}--document-preview-pdf-viewer {
+  position: relative;
+  @include pdfjsTextLayer;
```
jhpedemonte, line 5: Nice!

packages/discovery-styles/scss/components/document-preview/_pdfjs_web_mixins.scss
```diff
@@ -0,0 +1,83 @@
+@mixin pdfjsTextLayer {
```
jhpedemonte, line 1: So this is the file which is generated from `update-styles.sh`? If so, please add a comment stating as such, so that future maintainers know not to manually edit/update. Something like:

```css
/* DO NOT EDIT. THIS FILE IS AUTOMATICALLY GENERATED FROM `update-styles.sh`. */
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/useAsyncFunctionCall.ts
fukudasjp: I tried it but I found it needs some hack. So I choose this way to keep the React hook's way.

```typescript
// caller
useAsyncFunc(async (arg1,arg2) => {body;}, [arg1,arg2]);
// hook
function useAsyncCallback(asyncFunc, args) {
  useEffect(() => {
     // run await asyncFunc(...args);
     
     // disable hook dependency warning
  }, [asyncFunc, ...args]);
}
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
fukudasjp: Yes it is. It's to clear text spans under `textLayerDiv` rendered by previous text layer rendering.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/typings.d.ts
fukudasjp: The type definition is available on or after `pdfjs-dist@2.10.377` while ours is `2.2.228`. I can we upgrade the package?

packages/discovery-styles/scripts/update-styles.sh
fukudasjp: OK. I'd use CSS parser (`postcss`) and its AST to write the script.

packages/discovery-styles/package.json
fukudasjp: I didn't intend to run this script automatically. This script need to be run only when the `pdfjs-dist` package is upgraded.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/useAsyncFunctionCall.ts
jhpedemonte: Gotcha. Thanks for the explanation.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/typings.d.ts
jhpedemonte: Ah, ok. No, let's update `pdfjs-dist` in a separate PR. Also, we should be using a similar version for `@types/pdfjs-dist` as the installed `pdfjs-dist`.

packages/discovery-styles/scripts/generate-pdfjs_web_mixin.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Usage: node <this_script> <original pdfjs web css> <output scss>
+ */
+const postcss = require('postcss');
+const fs = require('fs');
+
+const originalPdfjsWebCss = process.argv[2];
+const mixinPdfjsWebScss = process.argv[3];
+
+// load teh original style
```
jhpedemonte, line 12:
```suggestion
// load the original style
```

packages/discovery-styles/scripts/update-styles.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+BASEDIR=$(dirname "$0")/..
+
+# pdfjs textLayer styles
+PDFJS_WEB_CSS=$BASEDIR/../../node_modules/pdfjs-dist/web/pdf_viewer.css
+PDFJS_SCSS=$BASEDIR/scss/components/document-preview/_pdfjs_web_mixins.scss
+node $BASEDIR/scripts/generate-pdfjs_web_mixin.js "$PDFJS_WEB_CSS" "$PDFJS_SCSS"
+
+# perttier
+../../node_modules/.bin/prettier --write "$PDFJS_SCSS"
```
jhpedemonte, line 19:
```suggestion
yarn run prettier --write "$PDFJS_SCSS"
```

jhpedemonte (approved):
Code is looking good. I have a few more suggestions, plus a few more from previous discussions (which I haven't closed as "resolved").

With those fixed, approved.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/typings.d.ts
fukudasjp: OK I update the version with `2.1.7`, which is the nearest version to the `pdfjs-dist@2.2.228`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
fukudasjp: OK. Let's think about it when we implement thumbnails. With that, our usage is only to display a (full-page) document and I drop the option.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
fukudasjp: Updated to use the same property both in `PdfViewer` and `PdfViewerTextLayer`

PR #238, opened 2021-11-10
feat: add PDF viewer with highlighting

fukudasjp:
This depends on #237.

This PR is to add `PDFViewerHighlighter` component which is a highlight layer for PDF based on text spans on fields in search result document. This also implements `PDFViewerWithHighlight` which is PDFViewerHighlighter integrated with PDFViewer.

The implementation includes a logic to find calculate bboxes for highlighting. Overview of the logic is in [README.md](https://github.com/watson-developer-cloud/discovery-components/blob/feat/highligh-on-pdf/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/README.md).

- Add generic utilities used by the logic
   - feat: add types and common utilities 45e5da8
- Add an option to extract HTML source in bbox from HTML field to processDoc
   - feat: add option for bbox text to processDoc c2ab1f6
- Implement the logic to calculate bboxes for highlighting (please refer to [README.md](https://github.com/watson-developer-cloud/discovery-components/blob/feat/highligh-on-pdf/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/README.md) to get high-level view of the logic)
   - feat: add text layer classes 20e4866
      - Implement `TextLayouts`
   - feat: add highlighting logic and README 839f948
      - Implement `textBoxMapping` and `Highlighter`
- Implement Rect components and storybook
  - feat: add PDF highlight component 67c0ac9

- Open the storybook **DocumentPreview > components > PdfViewerWithHighlight > with text selection**,
- In the right pane, select a field except for `html` from the dropdown on the top
- Select a range on the text
- Then, verify that the text on PDF is highlighted


jhpedemonte:
> How do you test/verify these changes?

Can you please add to this section?

---

Did a quick test in Storybook. I see the new `PdfViewerWithHighlight` section and I can select text on the right and see it appear on the left when selecting `text[0]` (I assume this is what it is supposed to show off). However, I don't see the same when selecting text in either `header[0]` or `footer[0]` -- should those work, too?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/CellProvider.ts
```diff
@@ -0,0 +1,96 @@
+import { isSideBySideOnLine } from '../common/bboxUtils';
+import { TextLayoutCell, TextLayoutCellBase } from '../textLayout/types';
+
+export class CellProvider {
+  private readonly skippedCells: TextLayoutCellBase[] = [];
+  private cells: TextLayoutCellBase[]; // make sure to handle this as immutable array
```
jhpedemonte, line 6: Then why not use `Object.freeze()` to actually make it immutable?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/CellProvider.ts
```diff
@@ -0,0 +1,96 @@
...
+        result.push(currentBox);
+        lastCell = baseCurrentCell;
+      }
+      lastCells = this.cells;
+      lastCursor = this.cursor;
+      lastResult = result;
+
+      return result;
+    };
+  })();
```
jhpedemonte, line 53: I'm confused by this setup: the function returns a function but it is immediately invoked? Why not just have a normal return?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/MappingSourceTextProvider.ts
```diff
@@ -0,0 +1,61 @@
...
+import minBy from 'lodash/minBy';
+import { spanGetText, spanLen, START } from '../common/textSpanUtils';
+import { TextLayoutCell } from '../textLayout/types';
+
+const debugOut = require('debug')?.('pdf:mapping:MappingSourceTextProvider');
+function debug(...args: any) {
+  debugOut?.apply(null, args);
+}
+
+export class MappingSourceTextProvider {
```
jhpedemonte, line 13: Reminder: the README does a good job of detailing the overall process. But each of these files should have a short description comment block as well.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/MappingTargetCellProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+    nextCellIndex: number;
+    normalizer: TextNormalizer;
+    leadingSpaces: number;
+  } | null = null;
+
+  constructor(cells: TextLayoutCellBase[]) {
+    this.cellProvider = new CellProvider(cells);
+  }
+
+  hasNext() {
```
jhpedemonte, line 18: I wonder: does it make sense to convert this and `MappingSourceTextProvider` to a [JS iterable](https://dev.to/xenoxdev/javascript-jungle-convert-any-object-to-iterable-40l6)?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textLayout/BaseTextLayout.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.id = id;
+    this.pageNum = pageNum;
+    this.bbox = bbox;
+    this.text = text;
+  }
+
+  getPartial(span: TextSpan): TextLayoutCellBase {
+    return new PartialTextLayoutCell(this, span);
+  }
+  getNormalized(): { cell: TextLayoutCell; span?: TextSpan } {
```
jhpedemonte, line 41: Nit: empty lines between these functions.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textLayout/PdfTextContentTextLayout.ts
```diff
@@ -0,0 +1,94 @@
...
+    return super.getBboxForTextSpan(span, options);
+  }
+
+  static getBbox(textItem: TextContentItem, viewport: PDFPageViewport): Bbox {
+    const { transform } = textItem;
+
+    const patchedViewport = viewport as PDFPageViewportOptions & PDFPageViewport;
+    const defaultSideways = patchedViewport.rotation % 180 !== 0;
+
+    // not sure this is true...
```
jhpedemonte, line 83: What else do you need to do here?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textLayout/dom.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  let left = cell.bbox[LEFT];
+  let right = cell.bbox[RIGHT];
+  const top = cell.bbox[TOP];
+  const bottom = cell.bbox[BOTTOM];
+
+  // convert offset
+  function getAdjustedOffset(orgOffset: number) {
+    return orgOffset;
+  }
```
jhpedemonte, line 34: This doesn't do anything.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textLayout/dom.ts
```diff
@@ -0,0 +1,67 @@
...
+        ? getTextNodeAndOffset(spanElement, getAdjustedOffset(endOffset))
+        : { textNode: spanElement.lastChild, textOffset: spanElement.lastChild.length };
+
+    debug('finding text node for: ', cell.text);
+    debug('  textContent: ', beginTextNode.textContent);
+    debug('  beginOffset: ', beginTextOffset);
+    debug('  textContent: ', endTextNode.textContent);
+    debug('    endOffset: ', endTextOffset);
+
+    const range = document.createRange();
```
jhpedemonte, line 51: 1. Does `range` need to be deleted/destroyed after it's done being used here? Or will the garbage collector take care of it?
2. This code has some big similarities to `createFieldRects` in `documentUtils.ts`. Should the be refactored to make use of same underlying code? Or would that be too complex?

jhpedemonte (commented):
One of the major things you were investigating was highlighting on Japanese text, correct? You should include a sample doc to show this off (similar to `__fixtures__/Art Effects.pdf`).

fukudasjp:
> Did a quick test in Storybook. I see the new `PdfViewerWithHighlight` section and I can select text on the right and see it appear on the left when selecting `text[0]` (I assume this is what it is supposed to show off). However, I don't see the same when selecting text in either `header[0]` or `footer[0]` -- should those work, too?

Thank you for pointing this out. They should work. I found some issue with them and will fix.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/TextNormalizer.ts
```diff
@@ -0,0 +1,266 @@
...
+    '‘', // U+2018
+    '‚', // U+201A
+    '‛', // U+201B
+    '❛', // U+275B
+    '❜', // U+275C
+    '❟' // U+275F
+  ].join('')}]`
+};
+
+const SURROGATE_PAIR = {
```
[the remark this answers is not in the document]
fukudasjp: It's a unicode surrogate pair -- a single letter but having length of `2` in javascript string. I add comment to the code.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/TextNormalizer.ts
```diff
@@ -0,0 +1,266 @@
...
+
+function normalizeDiacriticalMarks(text: string, keepLength = false) {
+  const r = text
+    .normalize('NFD')
+    .replace(DIACRITICAL_MARK_REGEX, DIACRITICAL_MARK.normal)
+    .normalize('NFC');
+  if (keepLength && r.length !== text.length) {
+    //
+    // String.normalize may change length of a string. `keepLength` flag keeps string
+    // length after conversion by padding or truncating a string.
```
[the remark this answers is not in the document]
fukudasjp: It rarely happen. The last step of getting the result string is `normalize('NFC')`. It basically shorten the original text. I added `substring` just in case.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/TextNormalizer.ts
```diff
@@ -0,0 +1,266 @@
...
+  if (spanLen(fromSpan) === spanLen(toSpan)) {
+    return toSpan[START] + (charIndex - fromSpan[START]);
+  }
+  return (
+    toSpan[START] +
+    Math.round((charIndex - fromSpan[START]) * (spanLen(toSpan) / spanLen(fromSpan)))
+  );
+}
+
+function optimizeSpanMappings(mappings: SpanMapping[]) {
```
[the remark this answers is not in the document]
fukudasjp: I will add a comment to this function.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/bboxUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Check whether two bbox intersect
+ *
+ * Same to `intersects` in DocumentPreview/utils/box.ts,
+ * but for type `Bbox`, which doesn't have page property
+ * @param boxA one bbox
+ * @param boxB another bbox
+ * @returns true iff boxA and boxB are overwrapped
+ */
+export function bboxIntersects(boxA: Bbox, boxB: Bbox) {
```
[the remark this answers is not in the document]
fukudasjp: Hmm. At least I should add return type to exported functions.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/bboxUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+  const [leftA, topA, rightA, bottomA] = boxA;
+  const [leftB, topB, rightB, bottomB] = boxB;
+  return !(leftB >= rightA || rightB <= leftA || topB >= bottomA || bottomB <= topA);
+}
+
+/**
+ * Get bbox for a text span assuming each character takes horizontal spaces evenly
```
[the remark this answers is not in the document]
fukudasjp: > What are the implications here if the characters are not evenly spaced?

In the case, left and right ends of highlight box don't exactly match to the beginning/ending characters of highlight span.

> And how often will we run into this scenario?

Currently, OCR text will be run into this case because the text won't be available as text object in PDF. I'm planning to improve it by implementing a logic to find bbox using DOM later.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/documentUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+export type ExtractedDocumentInfo = {
+  processedDoc: ProcessedDoc;
+  textMappings?: TextMappings;
+};
+
+export async function extractDocumentInfo(document: QueryResult) {
+  const docHtml = document.html;
+  const textMappings = getTextMappings(document) ?? undefined;
+
+  // HtmlView.tsx
```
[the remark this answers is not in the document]
fukudasjp: It should be just removed. Thanks

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/nonEmpty.ts
```diff
@@ -0,0 +1,3 @@
+export function nonEmpty<T>(value: T | null | undefined): value is T {
```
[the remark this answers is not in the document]
fukudasjp: > Add comment with summary of function.
👍 

> These files all need the copyright block.

I didn't see any typescript files have copyright comment in this repo. If we need it, could you suggest the comment content?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/textSpanUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!text) return text;
+  if (spanLen(span) === 0) return '';
+  return text.substring(span[START], span[END]);
+}
+
+export function spanLen(span: TextSpan) {
+  return Math.max(0, span[END] - span[START]);
+}
+
+export function spanIntersects([beginA, endA]: TextSpan, [beginB, endB]: TextSpan): boolean {
```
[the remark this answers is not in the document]
fukudasjp: > There already exists a function (with a more correct name) of spansIntersect in documentUtils.ts. Let's only have a single version.

I made the new method because of the difference of the type of arguments. I can refactor the new function to use the existing  `spansIntersect`. However, I just found the existing one is having off-by-one error and not sure the impact of fixing it. So, I tentatively added a TODO comment. Would you have any suggestion?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/CellProvider.ts
fukudasjp: :+1: I was't aware that I can use the type annotation `readonly TextLayoutCellBase[];`

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/CellProvider.ts
fukudasjp: As the method is expensive and can be called multiple times, I implemented a cache. The setup is to keep variables related to the cache in local scope. I agree it's not common and will rewrite it.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/MappingTargetCellProvider.ts
fukudasjp: Let me try :)

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textLayout/dom.ts
fukudasjp: > 1. Does range need to be deleted/destroyed after it's done being used here?

From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Range/detach), we used to use `Range.detach()`, but it does nothing in recent browsers.

fukudasjp: > 2. we can extract a common logic of iterating over rectangles on a range.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/bboxUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+  const [left, top, right, bottom] = bbox;
+  const width = right - left;
+  const resultLeft = left + (width / origLength) * spanStart;
+  const resultRight = left + (width / origLength) * spanEnd;
+
+  return [resultLeft, top, resultRight, bottom] as Bbox;
+}
+
+/**
+ * Check whether two bboxes seems to be side-by-side on a same line.
```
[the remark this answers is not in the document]
fukudasjp: Added comment to the code. This is used to get text on a row from a list of text layout cells. This determines whether to append text in a new cell to an existing text on a row or not.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/MappingTargetCellProvider.ts
fukudasjp: I tried, but I found that there is similarities between the class and JS iterable. But the class needs explicit user action `skip` or `consume` to move the _current index_ to the next. So, I'd like to keep this as-is.

We'll need `"downlevelIteration": true` in tsconfig.json to use user-defined iterator 😄

fukudasjp:
> One of the major things you were investigating was highlighting on Japanese text, correct? You should include a sample doc to show this off (similar to __fixtures__/Art Effects.pdf).

I just added a small Japanese PDF sample.

fukudasjp:
branch rebased on the latest `feat/redner-pdf-text` which includes yarn v2 updates.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/TextNormalizer.ts
```diff
@@ -0,0 +1,266 @@
+import { TextSpan } from '../../types';
+import { END, spanLen, START } from './textSpanUtils';
+
+type SpanMapping = { rawSpan: TextSpan; normalizedSpan: TextSpan };
+
+const SPACES = {
+  normal: () => ' ',
```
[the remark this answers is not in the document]
fukudasjp: fixed to make sure the function takes an argument, which is unused in this implementation.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/TextNormalizer.ts
```diff
@@ -0,0 +1,266 @@
...
+        normalizedText.length + newText.length
+      ];
+      normalizationMappings.push({ rawSpan, normalizedSpan });
+      addNormalizedText(newText);
+    }
+    this.normalizedText = normalizedText;
+    this.normalizationMappings = optimizeSpanMappings(normalizationMappings);
+  }
+
+  toNormalized(rawSpan: TextSpan): TextSpan {
```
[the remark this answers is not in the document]
fukudasjp: done.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/TextNormalizer.ts
```diff
@@ -0,0 +1,266 @@
...
+          normalizedText.length,
+          normalizedText.length + newText.length
+        ];
+        normalizationMappings.push({ rawSpan, normalizedSpan });
+        addNormalizedText(newText);
+        cur = re.lastIndex;
+      }
+      match = re.exec(this.rawText);
+    }
+    if (cur < this.rawText.length) {
```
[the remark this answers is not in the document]
fukudasjp: done.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/textSpanUtils.ts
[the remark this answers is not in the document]
fukudasjp: I tried to fix the existing `spansIntersect`, but it resulted in test error. Need to fix it outside of this PR and revisit. I added comment.

jhpedemonte:
1. The story shows selecting text and having it highlighted on the PDF. Does the reverse work? I'm thinking of scenario where user is presented with a PDF and selects some text and app needs to figure out which text-based enrichment it belongs to.
2. Have you tested scanned (non-programmatic) PDFs yet, which have gone through OCR? I'm wondering how well this code works in those cases.
    - Also as mentioned, we'll need to consider scanned documents which are skewed and how we handle highlighting there. I think the suggestion at the time (from Mauricio?) was that we may have enough info to rotate the original document so it doesn't appear skewed any more. In which case, highlighting shouldn't be an issue.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -0,0 +1,157 @@
...
+  highlights: DocumentFieldHighlight[];
+
+  /**
+   * PDF text content information in a page from parsed PDF
+   */
+  pdfRenderedText: PdfRenderedText | null;
+
+  /**
+   * Flag to whether or not to use bbox information from html field in the document.
+   * True by default. This is for testing and debugging purpose.
```
jhpedemonte, line 46: I think that we should better denote that these properties aren't meant to be used normally by end users. Best I can think is to prepend an underscore to the name:

```js
  _useHtmlBbox?: boolean;
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/nonEmpty.ts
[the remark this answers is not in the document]
jhpedemonte: > I didn't see any typescript files have copyright comment in this repo.

Huh, ok, we'll handle that in a separate PR.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerWithHighlight.stories.scss
```diff
@@ -0,0 +1,25 @@
+.withTextSelection {
+  display: flex;
+  height: 800px;
```
jhpedemonte, line 3: Why does this have an explicit height? Is it possible to make it relative?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerWithHighlight.stories.scss
```diff
@@ -0,0 +1,25 @@
...
+  .text {
+    overflow-wrap: break-word;
+    white-space: pre-wrap;
+    font-size: 10pt;
+    font-family: 'Courier New', Courier, monospace;
+  }
+
+  .highlight {
+    opacity: 0.4;
+    background: rgba(255, 64, 128, 1);
```
jhpedemonte, line 23: We should check what Carbon says about highlighting text and which colors/opacity to use.

jhpedemonte: Only thing I found so far is `$highlight` from https://www.carbondesignsystem.com/guidelines/color/usage. That's fine if there's a single highlight in the component, but we'll need more guidance if we have to handle multiple highlight colors.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/bboxUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+    return false;
+  }
+
+  const [leftA, topA, rightA, bottomA] = boxA;
+  const [leftB, topB, rightB, bottomB] = boxB;
+  const heightA = bottomA - topA;
+  const heightB = bottomB - topB;
+
+  // compare height ratio
+  const OVERLAP_RATIO = 0.8;
```
jhpedemonte, line 41: 1. Move this constant out of the function and to the top of the file.
2. Also, what is the purpose of this constance? How was it calculated? Add a description comment.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/documentUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+/**
+ * Extract bboxes and text_mappings from a search result document
+ */
+export async function extractDocumentInfo(document: QueryResult): Promise<ExtractedDocumentInfo> {
+  const docHtml = document.html;
+  const textMappings = getTextMappings(document) ?? undefined;
+
+  const processedDoc = await processDoc(
```
jhpedemonte, line 54: I have't fully mapped out the flow of the code, but is there a possibility of `processDoc` getting called multiple times for the same document? That would be expensive.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/nonEmpty.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * A filter to drop any non-null values from a list.
+ * Use with `Array.filter` method to get a list of non-null type.
+ *
+ * `const list: number[] = [1, null, 2].filter(nonEmpty); // [1,2]`
+ */
+export function nonEmpty<T>(value: T | null | undefined): value is T {
+  return value !== null && value !== undefined;
+}
```
jhpedemonte, line 9: This function is generic and not specific to `PdfViewerHighlight`. Move it up to the top-level `utils/`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/TextBoxMapping.ts
```diff
@@ -0,0 +1,116 @@
...
+import { TextNormalizer } from '../common/TextNormalizer';
+import { TextBoxMapping, TextBoxMappingEntry, TextBoxMappingResult } from './types';
+
+const debugOut = require('debug')?.('pdf:mapping:TextBoxMapping');
+function debug(...args: any) {
+  debugOut?.apply(null, args);
+}
+
+/**
+ * Text box mapping
```
jhpedemonte, line 21: This doesn't really explain this component. What does this map between?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/textBoxMapping/TextProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+  spanIncludesIndex,
+  spanGetText,
+  spanIntersection
+} from '../../../../utils/textSpan';
+import { findLargestIndex } from '../common/findLargestIndex';
+
+const MAX_HISTORY = 3;
+
+export type TextMatch = {
+  /** matched text span */
```
jhpedemonte, line 15: For single line comments like these, let's keep with convention of just starting with `//` and leave `/**` for multi-line comments (here and elsewhere in PR).

(Wonder if there's an eslint/prettier plugin for this?)

packages/discovery-react-components/src/components/DocumentPreview/utils/box.ts
```diff
@@ -1,16 +1,27 @@
 import { CellPage } from '../types';
 import { ProcessedBbox } from '../../../utils/document/processDoc';
 
+export const LEFT = 0;
+export const TOP = 1;
+export const RIGHT = 2;
+export const BOTTOM = 3;
```
jhpedemonte, line 7: Where are these used?

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
```diff
@@ -12,3 +12,16 @@
...
+  position: absolute;
+  transform-origin: left top 0px;
+  top: 0;
+  left: 0;
+}
+
+.#{$prefix}--document-preview-pdf-viewer-highlight--item {
+  position: absolute;
+  opacity: 0.5;
+  background: rgba(0, 0, 255, 1);
```
jhpedemonte, line 26: Same here: need to check with Design about use of highlight color/opacity.

jhpedemonte (approved):
Looks good overall and functions well in Storybook. Found a few minor things. With those fixed, approved.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerWithHighlight.stories.scss
fukudasjp: I update the code to use the `$highlight` color so far.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/documentUtils.ts
fukudasjp: This is expected to be called once for a document.

packages/discovery-styles/scss/components/document-preview/_document-preview-pdf-viewer.scss
fukudasjp: OK. Use the Carbon color with opacity as a default and check with design.

packages/discovery-react-components/src/components/DocumentPreview/utils/box.ts
fukudasjp: They were used in `dom.ts`. I refactored the code to use destructuring assignment.

fukudasjp:
> The story shows selecting text and having it highlighted on the PDF. Does the reverse work? I'm thinking of scenario where user is presented with a PDF and selects some text and app needs to figure out which text-based enrichment it belongs to.

Not yet, but I agree that it has to be considered to make this work as editor, too. I'm thinking to add a selection handler such as `onSelect({ field, fieldIndex, span })` to the component


> Have you tested scanned (non-programmatic) PDFs yet, which have gone through OCR? I'm wondering how well this code works in those cases.

I tried scanned document. Currently no highlight is shown on scanned document and need to fix it. On scanned PDF, highlight relies on bboxes in HTML field. With a document I tried, the bboxes spans on multiple lines and highlight were not accurate. Smaller bboxes are required for more accuracy.

PR #249, opened 2021-12-06
feat: document provider interface

jhpedemonte:

Allow consumer to specify an implementation of `DocumentProvider`, which will be used by `DocumentPreview` to query for the original document (PDF) to display. If not provided, continue previewing HTML or text.

Visit DocumentPreview in Storybook, for each sub-stories, check if PDF document type preview displays PDF. Note that highlighting is not ready in PDF.

In the Example Discovery search app, open a PDF named `Art Effects.pdf` (or you can choose any name in `ExampleDocumentProvider`), then see if the PDF is rendered correctly as DocumentPreview.

N/A

No


packages/discovery-react-components/src/components/DiscoverySearch/types.ts
```diff
@@ -11,3 +11,18 @@ export type SearchParams = Omit<DiscoveryV2.QueryParams, 'projectId' | 'headers'
...
+
+export interface DocumentProviderProps extends Pick<QueryResult, 'document_id' | 'metadata'> {
+  extracted_metadata?: JsonObject;
+}
+
+export interface DocumentProvider {
+  /**
+   * Return `true` if service can provide the given document
+   */
+  provides: (document: DocumentProviderProps) => boolean;
```
fukudasjp, line 23: I think we need to allow `async` way to provider functions. They may involve API calls.

jhpedemonte:
With latest code, I'm able to load a PDF (requires some CSS fixes, though!) in the example app:

Screen Shot 2021-12-07 at 4 05 16 PM

packages/discovery-react-components/package.json
```diff
@@ -61,6 +61,7 @@
     "@ibm-watson/discovery-styles": "1.x.x",
     "carbon-components": ">= 10.6.0 < 11",
     "carbon-components-react": ">= 7.6.0 < 8",
+    "core-js": "^2.6.12",
```
jhpedemonte, line 64: Example app was failing to properly load the `discovery-react-components` library, saying it couldn't find `core-js/es6/symbol`. This is due to `react-virtualized` still using `core-js` v2.x, which is no longer supported. I had to add `core-js` as a peer-dep here and as a dep for the example app.

Unfortunately, `react-virtualized` is no longer maintained and there are other of its dependencies which will cause us issues. We will have to find a suitable replacement, soon.

examples/discovery-search-app/src/App.js
```diff
@@ -75,7 +76,11 @@ const App = () => {
   ) : isError ? (
     <div>Unable to load Discovery projects. Please check your console for more details.</div>
   ) : (
-    <DiscoverySearch searchClient={searchClient} projectId={projectId}>
+    <DiscoverySearch
+      searchClient={searchClient}
+      projectId={projectId}
+      documentProvider={new ExampleDocumentProvider()}
```
jhpedemonte, line 82: This is how a user would enable the document provider, so that `DocumentPreview` invokes its methods automatically when wanting to render a document.

The other option is more "manual": simply manage the loading of `DocumentPreview` and pass in the correct `file` when needed.

examples/discovery-search-app/src/ExampleDocumentProvider.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * To see this in action, create a new collection in your project and upload
+ * "/examples/discovery-search-app/public/documents/Art Effects.pdf". Then, when running the
+ * example app against this collection, if you search and find this document, when you click
+ * "View passage in document" (or similar) to open the document view, you will see the PDF
+ * itself rendered, instead of HTML or text.
+ */
+export class ExampleDocumentProvider implements DocumentProvider {
+  async provides(document: DocumentProviderProps) {
+    return document.extracted_metadata.filename === 'Art Effects.pdf';
```
jhpedemonte, line 15: 🐛 One issue here is how will consumers of this library know which file to load? There exists a unique `document_id` for each document, but that's specific to Discovery and means nothing to end users.

Ideally, a consumer would be able to associate their own ID with the uploaded document, which Disco could then store with the document.

Any other ideas on how to handle this?

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -4,26 +4,28 @@ import { settings } from 'carbon-components';
...
 
 interface Props extends WithErrorBoundaryProps {
   /**
-   * Document data, as that returned by a query. Overrides result from SearchContext
+   * Document data, as that returned by a query. Overrides result from SearchContext.
    */
   document?: QueryResult;
   /**
-   * PDF file data as base64-encoded string
+   * PDF file data as "binary" string (array buffer). Overrides result from SearchContext.documentProvider.
```
jhpedemonte, line 23: Turns out previous comment was false. This `file` variable goes straight into `getDocument()` from `pdfjs-dist`, which takes an array buffer (not base64 encoded).

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -60,17 +65,61 @@ const DocumentPreview: FC<Props> = ({
...
+  const [pageCount, setPageCount] = useState(0);
+  const [pdfPageCount, setPdfPageCount] = useState(pageCount);
+  useEffect(() => {
+    if (providedFile && pdfPageCount > 0) {
+      setPageCount(pdfPageCount);
+    } else if (textMappings) {
+      const last = textMappings.text_mappings.length - 1;
+      setPageCount(textMappings?.text_mappings[last].page.page_number ?? 1);
+    }
+  }, [textMappings, providedFile, pdfPageCount]);
```
jhpedemonte, line 107: This code is only used by the `PdfViewer` -- it should be refactored to a better location in the code (maybe within `PdfViewer` itself.

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -60,17 +65,61 @@ const DocumentPreview: FC<Props> = ({
...
   const base = `${settings.prefix}--document-preview`;
 
   return (
     <div className={`${base}`}>
-      {(doc || file) && !didCatch ? (
+      {(doc || providedFile) && !didCatch ? (
         <>
+          {/* TODO This is only used when showing PDFs (currently). Should be moved into `PdfViewer` or
+                    elsewhere that makes more sense.
+          */}
```
jhpedemonte, line 117: @fukudasjp I think you were the one that pointed out the grey bar above all the files. This is it, the `PreviewToolbar`. I'm not sure how it fits in with your new redesign, but we should handle this better so it's only shown for PDFs, if that's only where it's needed.

packages/discovery-react-components/src/components/DocumentPreview/__stories__/DocumentPreview.stories.tsx
```diff
@@ -64,3 +98,26 @@ function passageSelection(doc: QueryResult, passages: object): QueryResult {
     document_passages: [passages[passageType]]
   };
 }
+
+function tableSelection(doc: QueryResult): QueryResult {
```
jhpedemonte, line 102: I added this back in as well, since it was there the last time we had the PDF view. However, it is not currently working. Needs to be investigated.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -19,7 +19,8 @@ type Props = PdfDisplayProps & {
   className?: string;
 
   /**
-   * PDF file data as base64-encoded string
+   * PDF file data as a "binary" string (array buffer)
+   * TODO Update to take `PDFSource` type (from pdfjs-dist) instead? Would allow binary data as well as URL.
```
jhpedemonte, line 23: The function `getDocument()` from `pdfjs-dist` can take much more than just an array buffer. In particular, it can also take a URL, which would be a nice feature for the `DocumentProvider` interface. We should update this and associated files with the expanded type.

For reference, here's the [type](https://github.com/mozilla/pdfjs-dist/blob/bd2a6cd87f9e0c3a9b3197d5d27acfb99886629b/types/src/display/api.d.ts#L618-L620) as defined in the latest `pdfjs-dist` version:

```
 * @param {string|URL|TypedArray|PDFDataRangeTransport|DocumentInitParameters}
 *   src - Can be a URL where a PDF file is located, a typed array (Uint8Array)
 *         already populated with data, or a parameter object.
```

The version we're using (v2.2.228) instead defines an object named `PDFSource` with `url` and `data` props. But if we're going to expand this type in our files, we should follow the lead of the latest `pdfjs-dist` code.

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -107,21 +160,44 @@ const DocumentPreview: FC<Props> = ({
...
   setLoading,
   hideToolbarControls,
   setHideToolbarControls,
   highlight
 }: PreviewDocumentProps): ReactElement | null {
+  // if we have PDF data, render that
+  // otherwise, render fallback document view
+  if (file) {
+    return (
+      <PdfViewer
```
jhpedemonte, line 190: @fukudasjp I think we can replace this with your `PdfViewerWithHighlight`, correct?

fukudasjp: Yes, but I recommend to do it in a separated PR because it will involve highlight location stuffs, too.

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/__tests__/CIDocument.spec.tsx
```diff
@@ -169,7 +169,7 @@ describe('<CIDocument />', () => {
 
     beforeEach(async () => {
       ({ getByTestId, getByText, findByText, findByTestId, queryByTestId } = render(
-        <CIDocument document={shortContract} overrideDocWidth={400} overrideDocHeight={600} />
+        <CIDocument document={shortContract} overrideDocWidth={400} overrideDocHeight={1200} />
```
fukudasjp, line 172: Upgrading `react-virtualized` from `9.21.1` to `9.21.2` or later makes this test unsuccessful.
```
src/components/CIDocument/components/CIDocument/__tests__/CIDocument.spec.tsx
...
  ● <CIDocument /> › Contract › loads correct document

    Unable to find an element with the text: On 22 December 2008 ART EFFECTS LIMITED and Customer. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```

On 9.21.1, with height `600`, entire the document is rendered. i.e. all virtual scroll items `VirtualScroll-8-0 to VirtualScroll-8-128` are rendered. 

But on 9.21.2, with height of `600`, only virtual scroll items `VirtualScroll-8-33 to VirtualScroll-8-128` are rendered. By increasing the height, we can render all the virtual scroll item.

I compared CIDocument between two versions, but I couldn't see difference. So, I update this test case to fix the CI error.

jhpedemonte: I vaguely remember running into issues when upgrading `react-virtualized`. For that reason, I kept it pegged specifically at `9.21.1` (and not something like `^9.21.1`).

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
fukudasjp: For future, we could be able to allows _page count_ even in HTML/TEXT preview. But for now, I agree it's only for PDF. So I move the code to `PdfViewer`.

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
fukudasjp: @jhpedemonte thanks for pointing this. I think about this toolbar and now I think that we can keep this toolbar code here. Because it's likely that the toolbar will have common action such as Search Text in document sooner or later. Instead, I will make the toolbar hidden when unnecessary.

fukudasjp:
For the styling issue in the example search app, PDFViewer's styles need to be updated. Let's handle it in a separated PR
image

examples/discovery-search-app/src/ExampleDocumentProvider.js
noah-eigenfeld: 🤔

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
fukudasjp: This is very good thing that we should consider. To align the latest `pdfjs-dist`, I'd add a new `src` prop and deprecate the `file` like below. However it wouldn't be small. So I think it'd good to leave this TODO for now and handle in a separated PR.

```
  /**
   * PDF data that PDFJS's getDocument can consume
   */
  src?: PDFSource; // Parameters<typeof PDFJSLib.getDocument>[0]
  /**
   * @deprecated PDF file data as a "binary" string (array buffer)
   * use `src={{data: <file prop value>}}` instead
   */
  file?: string;
```

noah-eigenfeld (approved):

noah-eigenfeld:
I tested in storybook, and these changes seemed to be working as intended

fukudasjp (approved):

PR #251, opened 2021-12-09
feat: update preview toolbar design for PDF viewer

fukudasjp:
Update preview toolbar design for updated PDF viewer
- Move pager to the left
- Add text description to zoom icons
- Allow user-defined actions

image

Use storybook (`DocumentPreview / components / PreviewToolbar / default`) and test the options using knobs.

N/A

No


jhpedemonte (approved):
Looks good. Only issue I have is that when clicking on the right-hand icons (like the zoom icons) the tooltip hangs around. Only way to make it go away is to click somewhere else -- I would have expected the tooltip to disappear if we no longer hover on that icon. But that can be handled in a separate PR.

PR #253, opened 2021-12-10
fix: fix PDF highlight misalignment issue

fukudasjp:
On PDFViewer with highlight, after you switch PDF pages, highlights on PDF gets misaligned.  This PR is to fix it.

In the screenshot, `documented acceptance` should be highlighted, but not in PDF view.
image

**Steps**
1. Open the storybook
2. Select the `documented acceptance` in the right pane. The text is highlighted on PDF
3. Move to page 2 using Knob
4. Move page back to 1
5. You see misaligned highlight

The issue comes from incorrect dependencies of a React hook. Also, the calculation of highlight rectangle starts before all the data is gathered.

Run the scenario using the storybook. Highlight shouldn't be misaligned even after moving pages back and force.

N/A

No


jhpedemonte (approved):

noah-eigenfeld (approved):
Code changes here look good. I also tested locally in storybook, and confirmed the expected behavior.

PR #259, opened 2022-01-11
feat: add "active" state to highlight on PdfViewer

fukudasjp:
Add "active" state to highlights on PdfViewer. When you set a highlight as _active_, 
- the highlight is displayed in darker color, and
- PdfViewer reveals the highlight by changing current page and scrolling
 
We can use Storybook (DocumentPreview > components > PdfViewerWithHighlight > default). In the story's Knobs tab, you can set **Highlights** and **Active highlight index** to the component.

no


jhpedemonte:
Functionality is good, but I'm concerned about the "active" color not meeting accessibility standards. It's hard to gauge since it's an overlay with opacity, but I used the **Digital Color Meter** app from the MacOS Developer Tools to measure both the overlay color and the text color (since the overlay changes the color of the underlying text):

Screen Shot 2022-01-19 at 4 02 35 PM
Screen Shot 2022-01-19 at 4 02 18 PM

I then ran those numbers through the [Contrast Checker](https://webaim.org/resources/contrastchecker/):

Screen Shot 2022-01-19 at 4 02 49 PM

If this were true system-level text highlighting, then the text would be true black, which easily passes all of the contrast checks. But since the overlay changes the text color, one test fails. We'll need to find a different color to use.

Or we can find a different way to denote "active"; for example, with a "focus" rectangle around it.

jhpedemonte:
If using the "Active highlight index" knob, should setting it to `2` force the PDF viewer to load the 3rd page, which contains this highlight index?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -54,40 +56,73 @@ const PdfViewerHighlight: FC<Props> = ({
...
+  return (
+    <div
+      ref={highlightDivRef}
+      className={cx(`${settings.prefix}--document-preview-pdf-viewer-highlight`, className)}
+    >
+      {highlightShapes.map((shape, hlIndex) => {
+        const active = shape?.highlightId ? activeIds?.includes(shape.highlightId) : false;
+        return (
+          <Highlight
+            key={shape?.highlightId || `highlight_${hlIndex}`}
```
jhpedemonte, line 78: How often will the "else" case resolve here? We shouldn't use array index for the React `key` -- that can easily lead to bugs.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -54,40 +56,73 @@ const PdfViewerHighlight: FC<Props> = ({
...
+}> = ({ className, activeClassName, shape, index, scale, active }) => {
+  function getPositionStyle(bbox: Bbox, scale: number, padding: number = 0) {
+    const [left, top, right, bottom] = bbox;
+    return {
+      left: `${(left - padding) * scale}px`,
+      top: `${(top - padding) * scale}px`,
+      width: `${(right - left + padding) * scale}px`,
+      height: `${(bottom - top + padding) * scale}px`
+    };
+  }
```
jhpedemonte, line 108: Move this function outside of the component, since it's standalone. No sense in redefining it each time the component renders.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -54,40 +56,73 @@ const PdfViewerHighlight: FC<Props> = ({
...
+
+  const highlightId = shape?.highlightId ?? `hl-${index}`;
   return (
-    <div className={cx(`${settings.prefix}--document-preview-pdf-viewer-highlight`, className)}>
-      {highlightBoxes.map((hl, hlIndex) => {
+    <div key={highlightId} data-highlight-id={highlightId}>
```
jhpedemonte, line 112: Same issue here with `key`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -54,40 +56,73 @@ const PdfViewerHighlight: FC<Props> = ({
...
-                    width: `${(right - left + padding) * scale}px`,
-                    height: `${(bottom - top + padding) * scale}px`
-                  }}
-                  data-testid="highlight"
-                />
-              );
-            })}
-          </React.Fragment>
+          <div
+            key={`hlId_${index}`}
```
jhpedemonte, line 116: And here.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -127,4 +162,33 @@ const useHighlighter = ({
   }, [document, isReady, pageNum, pdfRenderedText, processedDoc, textMappings]);
 };
 
+function useScrollIntoActiveHighlight(
+  highlightDivRef: React.MutableRefObject<HTMLDivElement | null>,
+  shapes: (HighlightShape | undefined)[],
```
jhpedemonte, line 167: `shapes` can be an array of `undefined`?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -127,4 +162,33 @@ const useHighlighter = ({
...
+    if (activeShape) {
+      const timer = setTimeout(() => {
+        const highlightDiv = highlightDivRef.current;
+        if (!highlightDiv) return;
+
+        const highlightElm = highlightDiv?.querySelector(
+          `[data-highlight-id=${activeShape.highlightId}]`
+        );
+        highlightElm?.firstElementChild?.scrollIntoView({ block: 'nearest', inline: 'nearest' });
+      }, 50);
```
jhpedemonte, line 187: Why `50` for the timeout? I don't like having arbitrary timeout values, since they may work on your computer, but may fail on a slower computer.

Using zero is fine since that's useful to make some code asynchronous. But it would be better to key off of something else rather than a random timeout, if possible.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
```diff
@@ -127,4 +162,33 @@ const useHighlighter = ({
...
+      return () => clearTimeout(timer);
```
jhpedemonte, line 188: If I'm reading this right, this is meant to be the cleanup function for the `useEffect`, right? If so, please comment.

Or better yet, do it like they show in the [docs](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1):

```
return function cleanup() {
  clearTimeout(timer);
}
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerWithHighlight.tsx
```diff
@@ -13,35 +18,126 @@ type Props = PdfViewerProps & HighlightProps;
...
+ */
+function useMovePageToActiveHighlight(
+  page: number,
+  activeHighlightPages: (number | null)[],
+  setPage: ((page: number) => any) | undefined
+) {
+  const [highlightPage, setHighlightPage] = useState<number | undefined>();
+
+  useEffect(() => {
+    const pages = activeHighlightPages.filter(nonEmpty);
```
jhpedemonte, line 75: Wouldn't it make more sense to move this into `useActiveHighlightPages`?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerWithHighlight.tsx
```diff
@@ -13,35 +18,126 @@ type Props = PdfViewerProps & HighlightProps;
...
+    if (pages.length === 0 || pages.includes(page)) {
+      return;
+    }
+    const newPage = pages[0];
+    setHighlightPage(newPage);
+  }, [page, activeHighlightPages]);
+
+  const [currentPage, setCurrentPage] = useState(page);
```
jhpedemonte, line 83: I think this function would read better if this were moved to the top and it referenced `currentPage` instead of `page` in the rest of the code.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/documentUtils.ts
```diff
@@ -20,17 +20,20 @@ export function getDocFieldValue(
...
+  let fieldText: string | object | undefined;
 
   const documentFieldArray = document[field];
   if (!Array.isArray(documentFieldArray) && !index) {
     fieldText = documentFieldArray;
   } else {
     fieldText = documentFieldArray?.[index ?? 0];
   }
-  if (!fieldText || !span) {
+  if (typeof fieldText?.['table_text'] === 'string') {
```
jhpedemonte, line 31: Why this change?

jhpedemonte (commented):
I keep getting confused between `PdfViewerHighlight.tsx` and `PdfViewerWithHighlight.tsx`. I think one of those should be renamed, at least, to make things clearer.

fukudasjp:
@jhpedemonte thanks for heading up [the contrast issue](https://github.com/watson-developer-cloud/discovery-components/pull/259#issuecomment-1016918989)! I updated the highlight style to use smaller opacity value. That passes the test.

image

fukudasjp:
> Oh, after refreshing Storybook, it worked. But then going back to 1 and 0 did nothing.

This issue came from the Storybook code and fixed it.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
fukudasjp: The `shape.highlightId` is currently optional, but I refactor the code to make the `highlightId` required.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
fukudasjp: no. I fixed the code

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerHighlight.tsx
fukudasjp: The intention was to push a new task to the javascript runtime. So I use the value `0`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/utils/common/documentUtils.ts
fukudasjp: Highlighting went wrong when there is `table` field in a document because it contains an object instead of text. Also, there is text_mappings for between bboxes and text in `table` fields. That code is to specially handles the `table` field to get field text referred in the text_mappings.

As the code is not clear, I updated the code and added a comment.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerHighlight/PdfViewerWithHighlight.tsx
fukudasjp: Moved the `currentPage useState` to the top, revised the code and added comments.

fukudasjp:
> I keep getting confused between PdfViewerHighlight.tsx and PdfViewerWithHighlight.tsx. I think one of those should be renamed, at least, to make things clearer.

I agree. Just renaming PdfViewerHighlight to PdfHighlight makes the situation much better and it's the first step, I think. I drafted  #275 on top of this PR.

jhpedemonte (approved):

PR #261, opened 2022-01-14
chore: refactor to make util function to detect preview type

fukudasjp:
Made an util function to detect the document preview type, which will be used when choosing what component should be used for rendering. This is to easily ensure preview type described in the ADR.

- Making unit test
- Manual test

n/a

no


fukudasjp:
Taken over by #294

PR #283, opened 2022-01-27
fix: add missing prop to doc provider interface

jhpedemonte:

Add `result_metadata` prop type to `DocumentProviderProps` interface


fukudasjp (approved):
LGTM.

Not directly related to this PR, but some fixtures (e.g. [multiPageResults.json](https://github.com/watson-developer-cloud/discovery-components/blob/master/examples/discovery-search-app/cypress/fixtures/query/multiPageResults.json)) for query results don't have the `result_metadata` property even the property is marked as required in the type definition. 
Maybe we have to fix the fixtures later.

PR #290, opened 2022-02-03
chore: update vulnerable deps

jhpedemonte:

Fix many of the vulnerable deps found in https://github.com/watson-developer-cloud/discovery-components/security/dependabot?q=is%3Aopen+sort%3Aseverity.


ColbyJohnIBM (approved):
This just a `yarn.lock` change? Does that just mean `yarn` should have been run before previous commits?

jhpedemonte:
> Does that just mean `yarn` should have been run before previous commits?

No. These are dependencies of dependencies, so they only appear in `yarn.lock`, where their versions are set. I used [`yarn set resolution`](https://yarnpkg.com/cli/set/resolution) to update the versions within the lockfile to non-vulnerable versions.

PR #306, opened 2022-02-21
feat: add PDFSource support to PDFViewer and DocumentProvider

fukudasjp:

Allow `PDFSource` as a document source of `PDFViewer` and `DocumentProvider` to exposing various options such as `cMapUrl` defined in `PDFSource`.

Specifically, (1) `file` property of `PDFViewer` can take `PDFSource` type and (2) `get` method of `DocumentProvider` can return an object including `PDFSource`. 

- Choose `PDF source type` in storybook DocumentPreview/components/PDFViewer/default
- Author tested `DocumentProvider#get`

no


packages/discovery-react-components/src/components/DiscoverySearch/types.ts
```diff
@@ -17,6 +18,14 @@ export interface DocumentProviderProps
   extracted_metadata?: JsonObject;
 }
 
+/**
+ * Document data for PDFSource
+ */
+export type PDFSourceDocument = {
+  __type: 'pdf';
+  source: PDFSource;
+};
```
fukudasjp, line 27: Instead of simply allowing `PDFSource` object, use an object with `__type` so that we'd be able to support various type of documents in future.

packages/discovery-react-components/src/components/DiscoverySearch/types.ts
```diff
@@ -17,6 +18,14 @@ export interface DocumentProviderProps
   extracted_metadata?: JsonObject;
 }
 
+/**
+ * Document data for PDFSource
+ */
+export type PDFSourceDocument = {
+  __type: 'pdf';
```
jhpedemonte, line 25: Is there a reason to use `__`? I prefer simply:

```suggestion
   type: 'pdf';
```

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -82,12 +84,17 @@ const DocumentPreview: FC<Props> = ({
...
+          setProvidedFile((file as PDFSourceDocument).source);
+        } else {
+          setProvidedFile(undefined);
+        }
       }
     }
 
     // `file` takes precedence over a file provided by `documentProvider`
-    if (file) {
+    if (typeof file === 'string') {
```
jhpedemonte, line 97: We should expand the `file` prop of `DocumentPreview` to also accept `PDFSource`. That way, a consumer of `DocumentPreview` can make use of that (for example, by simply passing in a URL instead of first donwloading all the data).

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -82,12 +84,17 @@ const DocumentPreview: FC<Props> = ({
     async function fetchFile() {
       const hasFile = await documentProvider!.provides(document!);
       if (hasFile) {
-        setProvidedFile(await documentProvider?.get(document!));
+        const file = await documentProvider?.get(document!);
+        if (file?.['__type'] === 'pdf') {
```
jhpedemonte, line 88:
```suggestion
        if (file?.__type === 'pdf') {
```

packages/discovery-react-components/src/components/DiscoverySearch/types.ts
```diff
@@ -25,5 +34,5 @@ export interface DocumentProvider {
   /**
    * Return document data for given document, as a "binary" string (array buffer)
    */
```
jhpedemonte, line 36: Update comment to match.

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -149,8 +156,8 @@ const DocumentPreview: FC<Props> = ({
   );
 };
 
-interface PreviewDocumentProps
-  extends Pick<Props, 'document' | 'file' | 'highlight' | 'fallbackComponent'> {
+interface PreviewDocumentProps extends Pick<Props, 'document' | 'highlight' | 'fallbackComponent'> {
+  file?: string | PDFSource;
```
jhpedemonte, line 160: This change wouldn't be necessary if we updated `DocumentPreview#file`.

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -42,7 +43,10 @@ export function isJsonFile(doc: QueryResult | null | undefined): boolean {
 /**
  * Returns the preview type for document
  */
-export function detectPreviewType(document: DiscoveryDocument, file?: string): PreviewType {
+export function detectPreviewType(
+  document: DiscoveryDocument,
+  file?: string | PDFSource
```
jhpedemonte, line 48: We are often repeating `string | PDFSource` type. We should alias that to a type named `DocumentFile` or something like that. Then, later, if we add another supported type, we only need to update the definition.

jhpedemonte (approved):
Looks good. Have some minor suggestions. With those fixed, approved.

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
fukudasjp: Thank you for pointing this out. It made the code much clean. As we accept `ArrayBuffer` as well, the type is `string | ArrayBuffer | PDFSource`

PR #313, opened 2022-02-27
fix: fix and improve PDF highlighting

fukudasjp:
#### What do these changes do/fix?
- fix redundant DOM elements https://github.com/watson-developer-cloud/discovery-components/pull/313#discussion_r815440688
- fix text mapping fallback https://github.com/watson-developer-cloud/discovery-components/pull/313#discussion_r817473858
- fix fragmented highlights https://github.com/watson-developer-cloud/discovery-components/pull/313#discussion_r815441941
- fix PDF text mappings https://github.com/watson-developer-cloud/discovery-components/pull/313#discussion_r815442587
- improve PDF text mappings by giving priority to longer matches. So, updated the code to repeat text match for the minimum length of 27, 9, 3, 1. Added `minLength` parameter to many of text match functions. https://github.com/watson-developer-cloud/discovery-components/pull/313#discussion_r817564062

#### How do you test/verify these changes?
Please try the storybook and see if PDF highlights works as before.

#### Have you documented your changes (if necessary)?
n/a

#### Are there any breaking changes included in this pull request?
no

PR #319, opened 2022-03-07
fix: fix PDF highlighting

fukudasjp:

Follow up #313 and improve PDF highlighting accuracy.

- Fix sideeffect https://github.com/watson-developer-cloud/discovery-components/pull/313#issuecomment-1058893548
   - This is from an issue in finding a text span in `document.text` of a text from a PDF text content. The logic should have found the text within a range, but the logic tried to find the text from entire the `document.text`. So, the highlight location was broken. So, fix the logic to respect the range to search text. image
- Normalize null character in PDF text content
   - PDF in some languages such as Serbian and Hindi includes null characters `\x00` in its text contents. They break PDF highlighting as the null characters are not extracted to document's field. So, normalize `\x00` to an empty string. image

- I added a testcase as the issues highly depend on PDF file and document
- ad-hoc tests in storybook


packages/discovery-react-components/src/components/DocumentPreview/components/PdfHighlight/utils/textBoxMapping/TextProvider.ts
```diff
@@ -50,14 +51,30 @@ export class TextProvider {
...
+   *   searchSpan: the span where the `text` is searched
    */
-  getMatches(text: string, minLength = 1, maxLength = text.length): TextMatch[] {
+  getMatches(
+    text: string,
+    {
+      minLength = 1,
+      maxLength = text.length,
+      searchSpan
+    }: { minLength?: number; maxLength?: number; searchSpan?: TextSpan } = {}
```
jhpedemonte, line 65: Instead of having a positional arg and then an "options" object, why not just make a single arguments object with everything:
```suggestion
    { 
      text,
      minLength = 1,
      maxLength = text.length,
      searchSpan
    }: { 
      text: string;
      minLength?: number;
      maxLength?: number;
      searchSpan?: TextSpan
    } = {}
```

Here and other implementations of `getMatches()`

jhpedemonte (approved):
Looks good and thanks for adding detailed tests. Just had one small nit.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfHighlight/utils/textBoxMapping/TextProvider.ts
fukudasjp: Taking `text` argument is because it's more important than others. Others are options.

PR #320, opened 2022-03-10
feat: add state change event to DocumentPreview component

fukudasjp:
This PR is to add `onPreviewStateChange` event to the `DocumentPreview` component so that its parent component can be notified with the preview state changes.

I added an storybook action to the DocumentPreview storybooks. Switch to the Actions tab in the bottom pane and see if the state change event is logged.
image

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
@@ -54,6 +54,10 @@ interface Props extends WithErrorBoundaryProps {
    * React component rendered as a fallback when no preview is available
    */
   fallbackComponent?: ComponentProps<typeof SimpleDocument>['fallbackComponent'];
+  /**
+   * Callback to receive changes in document preview state
+   */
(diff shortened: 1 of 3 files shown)


packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -54,6 +54,10 @@ interface Props extends WithErrorBoundaryProps {
    * React component rendered as a fallback when no preview is available
    */
   fallbackComponent?: ComponentProps<typeof SimpleDocument>['fallbackComponent'];
+  /**
+   * Callback to receive changes in document preview state
+   */
+  onPreviewStateChange?: (state: { currentPage?: number }) => void;
```
fukudasjp, line 60: I choose generic name rather than `onCurrentPageChange` so that we can easily add events to be notified.

jhpedemonte: In that case, should we just call this `onChange`? Using "state" is a bit redundant, and we already know it's the "preview" component.

jhpedemonte (approved):

PR #322, opened 2022-03-11
fix: show text view if there is no text_mappings

wataken-ibm:
Changed to show the TEXT preview if the document does not have `text_mappings` field instead of HTML preview

Update unit tests

N/A

No


packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -58,7 +58,7 @@ export function detectPreviewType(document: DiscoveryDocument, file?: DocumentFi
 
   const isJsonType = isJsonFile(document);
   const isCsvType = isCsvFile(document);
-  if (document.html && !isJsonType && !isCsvType) {
+  if (document.html && hasTextMappings && !isJsonType && !isCsvType) {
```
fukudasjp, line 61: I think that we can still render `HTML` when we don't have `passage`. That's similar to PDF.

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -59,7 +59,9 @@ export function detectPreviewType(document: DiscoveryDocument, file?: DocumentFi
   const isJsonType = isJsonFile(document);
   const isCsvType = isCsvFile(document);
   if (document.html && !isJsonType && !isCsvType) {
-    return 'HTML';
+    if (hasTextMappings || !hasPassage) {
```
jhpedemonte, line 62: Please add a comment here about _why_ this code is like this, since it is not obvious just from the code.

jhpedemonte (approved):

PR #323, opened 2022-03-11
fix: fix page switching issue with PdfViewerWithHighlight

fukudasjp:
This PR is to fix the following issues related to page switching in PdfViewer with highlights.

- When PDF viewer is initially shown with active highlight, the first page change (for example,  by user's clicking on Next page button) is ignored
   - Fixed by improving logic of page change. The condition of setting page was wrong.
- Suppose that PDF viewer is initially shown with active highlight, and user moves the current page. When user tries to activate another highlight on the same page to the initially activated highlight, the newly activated highlight is not revealed. Instead, PDF viewer keep showing the same page. 
    - The internal logic tracked only page number of an active highlight. So it couldn't detect changes of active highlight  when the highlights are on the same page. Now the logic tracks active highlight ID as well to properly detect the changes.

I added a new storybook. I used the storybook to run the problematic scenario.
image

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
@@ -171,7 +171,7 @@ function useHighlightState({
 /**
  * Hook to move PDF page depending on active highlight
  */
-function useMovePageToActiveHighlight(
+export function useMovePageToActiveHighlight(
   page: number,
   activeHighlightPages: number[],
   activeIds: string[] | undefined,
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
@@ -25,6 +25,7 @@ import {
 } from '../PdfHighlight/utils/common/highlightUtils';
 import flatMap from 'lodash/flatMap';
 import uniq from 'lodash/uniq';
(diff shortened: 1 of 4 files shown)


packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
```diff
@@ -165,37 +171,41 @@ function useHighlightState({
...
+        !isEqual(prevHighlightRef.current, { activeHighlightPages, activeIds }) && // active highlight is changed
+        !activeHighlightPages.includes(currentPage) // `currentPage` doesn't show active highlight page
+      ) {
+        const highlightPage = activeHighlightPages[0];
+        setCurrentPage(highlightPage);
+        setPage?.(highlightPage);
+      }
     }
-  }, [activeHighlightPages, currentPage, setPage]);
+    prevHighlightRef.current = { activeHighlightPages, activeIds };
```
fukudasjp, line 207: We should have updated this ref value always. Early return and conditional update in the previous code made an issue.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
```diff
@@ -165,37 +171,41 @@ function useHighlightState({
 /**
  * Hook to move PDF page depending on active highlight
  */
-function useMovePageToActiveHighlight(
+export function useMovePageToActiveHighlight(
   page: number,
   activeHighlightPages: number[],
+  activeIds: string[] | undefined,
```
jhpedemonte, line 177: I think you can just make this optional instead of using `undefined`:
```suggestion
  activeIds?: string[],
```

jhpedemonte (approved):

PR #332, opened 2022-04-13
feat: limit response size

jhpedemonte:

Update `SearchResults` component to update the search parameters and limit the size of the response, to only what is needed to render `SearchResults`.
Update `fetchDocuments()` to take collection ID argument. Fix how response is merged into search store response.
Fix some eslint warnings.

Run example app and run a blank search. Note the size of the `/query` response before and after. Ensure that loading a document shows all document data.

Yes. See comments below.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,6 +11,12 @@
   },
   "rules": {
     "prettier/prettier": "error",
-    "no-only-tests/no-only-tests": "error"
+    "no-only-tests/no-only-tests": "error",
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "(useDeepCompareMemo|useDeepCompareCallback|useDeepCompareEffect)"
+      }
+    ]
   }
 }
--- a/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
+++ b/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
@@ -353,7 +353,7 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
 
   useDeepCompareEffect(() => {
     setSearchResponse(overrideSearchResults);
-  }, [overrideSearchResults]);
+  }, [overrideSearchResults, setSearchResponse]);
 
   useDeepCompareEffect(() => {
     setSearchParameters(currentSearchParameters => {
@@ -363,11 +363,11 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
         ...deprecateReturnFields(overrideQueryParameters)
       };
     });
-  }, [projectId, overrideQueryParameters]);
+  }, [projectId, overrideQueryParameters, setSearchParameters]);
 
   useDeepCompareEffect(() => {
     setGlobalAggregationsResponse(overrideAggregationResults);
-  }, [overrideAggregationResults]);
+  }, [overrideAggregationResults, setGlobalAggregationsResponse]);
 
   useDeepCompareEffect(() => {
     setCollectionsResults(overrideCollectionsResults);
@@ -379,7 +379,7 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
 
   useDeepCompareEffect(() => {
     setAutocompletions(overrideAutocompletionResults);
-  }, [overrideAutocompletionResults]);
+  }, [overrideAutocompletionResults, setAutocompletions]);
 
   useDeepCompareEffect(() => {
     setComponentSettings(overrideComponentSettings);
@@ -440,7 +440,7 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
         setAutocompletions(undefined);
       }
     },
-    [autocompletionOptions, projectId]
+    [autocompletionOptions, fetchAutocompletions, projectId, setAutocompletions]
   );
 
   const handleFetchAggregations = useCallback(
--- a/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
+++ b/packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
@@ -122,7 +122,11 @@ export interface SearchApiIFC {
   performSearch: (searchParameters: DiscoveryV2.QueryParams, resetAggregations?: boolean) => void;
   fetchAutocompletions: (nlq: string) => Promise<void>;
   fetchAggregations: (searchParameters: DiscoveryV2.QueryParams) => Promise<void>;
-  fetchDocuments: (filterString: string, searchResponse: DiscoveryV2.QueryResponse | null) => void;
+  fetchDocuments: (
+    filterString: string,
+    collections?: string[],
+    searchResponse?: DiscoveryV2.QueryResponse
+  ) => void;
   setSelectedResult: (result: SelectedResult) => void;
   setAutocompletionOptions: (
     autoCompletionOptions: AutocompletionOptions | React.SetStateAction<AutocompletionOptions>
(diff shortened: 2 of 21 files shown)


.eslintrc
```diff
@@ -11,6 +11,12 @@
   },
   "rules": {
     "prettier/prettier": "error",
-    "no-only-tests/no-only-tests": "error"
+    "no-only-tests/no-only-tests": "error",
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "(useDeepCompareMemo|useDeepCompareCallback|useDeepCompareEffect)"
```
jhpedemonte, line 18: Get ESLint warnings for our custom `useEffect` hooks.

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -122,7 +122,11 @@ export interface SearchApiIFC {
   performSearch: (searchParameters: DiscoveryV2.QueryParams, resetAggregations?: boolean) => void;
   fetchAutocompletions: (nlq: string) => Promise<void>;
   fetchAggregations: (searchParameters: DiscoveryV2.QueryParams) => Promise<void>;
-  fetchDocuments: (filterString: string, searchResponse: DiscoveryV2.QueryResponse | null) => void;
+  fetchDocuments: (
+    filterString: string,
+    collections?: string[],
+    searchResponse?: DiscoveryV2.QueryResponse
+  ) => void;
```
jhpedemonte, line 129: This is a BREAKING CHANGE: we need to specify a collection ID as well, when fetching documents based on ID, since doc IDs are only unique within collections, not projects.

jhpedemonte, line 129: This is a BREAKING CHANGE: we need to specify a collection ID as well, when fetching documents based on ID, since doc IDs are only unique within collections, not projects.

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -480,14 +484,46 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
+                ...results[idx],
+                ...filteredResult
+              };
+              results.splice(idx, 1, mergedResult);
+            } else {
+              results.push(filteredResult);
+            }
+          }
+
           setSearchResponse({
```
jhpedemonte, line 523: For some reason that I don't know, `fetchDocument()` updates the search response store.

There are several "stores" handled by this code, two of which are:
- the `searchResponseStore`, which is updated by `performSearch()`
- the `fetchDocumentsResponseStore`, which is updated by `fetchDocument()`

But `fetchDocument()` also updates `searchResponseStore`. Why?

Best I can figure is something like the scenario that this PR enables: perform a search but only request the minimal data required (which ends up in the `searchResponseStore`). But then we want to display then entire document, so we call `fetchDocument()` which returns the full doc data. So may as well add this new data to the existing one in `searchResponseStore`.

But I think that should be an option for `fetchDocument()`, not the default. That is, by default, calling the method will only update the `fetchDocumentsResponseStore`. If you also want to update the `searchResponseStore`, the pass a boolean option. Thoughts?

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
```diff
@@ -480,14 +484,46 @@ const DiscoverySearch: FC<DiscoverySearchProps> = ({
...
             ...searchResponse,
-            matching_results:
-              (searchResponse.matching_results as number) + filteredResponse.results.length,
-            results: [...searchResponse.results, ...filteredResponse.results]
+            matching_results: results.length,
+            results
```
jhpedemonte, line 526: The old code simply appended the new doc data to the existing `searchResponseStore` data. But this lead to duplicated documents, as well as more items than were there previously.

The code above instead will merge the document data with any existing data, assuming that the doc and collection IDs match. Any unmatched docs are still appended.

This could be considered a BREAKING CHANGE, but I think this is actually fixing a bug in the previous implementation.

packages/discovery-react-components/src/components/SearchResults/SearchResults.tsx
```diff
@@ -125,6 +139,8 @@ const SearchResults: React.FunctionComponent<SearchResultsProps> = ({
     typeof showTablesOnlyToggle === 'undefined' ? hasTables : showTablesOnlyToggle
   );
 
+  useUpdateQueryReturnParam({ displaySettings, resultLinkField });
```
jhpedemonte, line 142: This is another BREAKING CHANGE, since we are changing what the parent context requests when doing a search.

jhpedemonte:
Quick test:
- query Sample Project in Tooling with 50 count: 2.4MB 13.36s
- query same Sample Project in example app with 50 count: 1.1 MB 5.33s

packages/discovery-react-components/src/components/SearchResults/SearchResults.tsx
```diff
@@ -256,6 +273,45 @@ const SearchResults: React.FunctionComponent<SearchResultsProps> = ({
   );
 };
 
+/**
+ * Hook to update search parameters' `return` param with the minimal necessary fields
+ * that are necessary to render this component.
```
noah-eigenfeld, line 278: Confused a bit by the wording of this comment, before I go read the actual code. Is this closer to a reset-to-defaults hook? Or are there some parts of the search parameters that will remain the same, meanwhile others are getting trimmed off?

packages/discovery-react-components/src/components/SearchResults/SearchResults.tsx
```diff
@@ -256,6 +273,45 @@ const SearchResults: React.FunctionComponent<SearchResultsProps> = ({
...
+ *
+ * NOTE: This hook must be specified before any other use of `setSearchParameters`.
```
noah-eigenfeld, line 280: Should we create a hook that does both and start using that from now on?

noah-eigenfeld: Re-reading the comment, and I'm wondering if my initial read was what you meant. Does this have to run once and only once before using `setSearchParameters` or before each and every use of `setSearchParameters`?

noah-eigenfeld (commented):
A few comments. Started testing, but realized I need to use a project with more files to get a better look at the response sizes. Will pick it back up tomorrow. So far, code looks alright.

packages/discovery-react-components/src/components/SearchResults/SearchResults.tsx
jhpedemonte: This hook will update search params any time they change. I went ahead and removed that note, though, since I don't think that's true any more.

noah-eigenfeld (approved):
Tested locally by linking to the related tooling PR. Saw decrease in query size.
Before:
Screen Shot 2022-04-21 at 1 47 40 PM

After:
Screen Shot 2022-04-21 at 1 47 48 PM

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
```diff
@@ -118,7 +118,9 @@ export const MultiSelectFacetsGroup: FC<MultiSelectFacetsGroupProps> = ({
           tempSelectedFacets.find(({ selectedFacetKey, checked }) => {
             if (selectedFacetKey === facetText) {
               facetSelected = !!checked;
+              return true;
             }
+            return false;
```
ColbyJohnIBM, line 123: Why are there return statements needed? Doesn't seem like the result of the `find` is used anywhere...

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
ColbyJohnIBM: Do these breaking changes mean we should do a major release version bump?

ColbyJohnIBM:
Code looks good, pulling down and checking locally.

ColbyJohnIBM (approved):
Code looks good and runs well locally-- I'm seeing similar drastic reductions in the response size (90+% reductions in some cases).

packages/discovery-react-components/src/components/DiscoverySearch/DiscoverySearch.tsx
jhpedemonte: Theoretically, including the words `BREAKING CHANGE` in the commit should make it so that our release code automatically updates the major version. If it doesn't, I'll do it manually.

packages/discovery-react-components/src/components/SearchFacets/components/FacetsGroups/MultiSelectFacetsGroup.tsx
jhpedemonte: 1. This change was made due to linter saying we weren't using `find()` correctly.
2. But now I see that this should instead be a `forEach`, since we're not using the result.

jhpedemonte: Actually, changing to use `every`, since that can be short-circuited once we have a match.

PR #340, opened 2022-04-26
fix: Mitigate some pagination issues in search results

ColbyJohnIBM:

- Only include each document once in filter string
- Allow pagination to be reset externally

##### BREAKING CHANGES:
- No longer display table-only results with passage results
- No longer add page size to pagination options (if it wasn't already included)
- Use count values exclusively (ignore results_per_page)

Contributes to [Watson-Discovery/disco-issue-tracker#4145](https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/4145)

Run example app or against tooling

TO BE ADDED

Yes, see above.

### DIFF
--- a/packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
+++ b/packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
@@ -1,5 +1,15 @@
-import React, { FC, useContext, useState, useEffect } from 'react';
+import React, {
+  FC,
+  useContext,
+  useState,
+  useEffect,
+  useCallback,
+  useImperativeHandle,
+  forwardRef,
+  ForwardedRef
+} from 'react';
 import { unstable_Pagination as CarbonPagination } from 'carbon-components-react';
+import cx from 'classnames';
 import { SearchApi, SearchContext } from 'components/DiscoverySearch/DiscoverySearch';
 import DiscoveryV2 from 'ibm-watson/discovery/v2';
 import { settings } from 'carbon-components';
@@ -35,6 +45,10 @@ export interface ResultsPaginationProps {
    */
   onChange?: (e: ResultsPaginationEvent) => void;
   /**
+   * ref passed in by parent that can be used to reset Carbon's pagination component
+   */
+  resetRef?: ForwardedRef<any>;
+  /**
    * Additional props to be passed into Carbon's Pagination component
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -46,6 +60,10 @@ interface ResultsPaginationEvent {
   pageSize: number;
 }
 
+interface ResetOptions {
+  triggerOnChange?: boolean;
+}
+
 const ResultsPagination: FC<ResultsPaginationProps> = ({
   page = 1,
   pageSizes = [10, 20, 30, 40, 50],
@@ -53,6 +71,7 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
   showPageSizeSelector = true,
   messages = defaultMessages,
   onChange,
+  resetRef,
   ...inputProps
 }) => {
   const mergedMessages = { ...defaultMessages, ...messages };
@@ -63,7 +82,47 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
     isResultsPaginationComponentHidden
   } = useContext(SearchContext);
   const [currentPage, setCurrentPage] = useState(page);
-  const resultsPerPage = componentSettings?.results_per_page ?? 10;
+  const [resetCounter, setResetCounter] = useState(0);
+  const resultsPerPage = pageSize || searchParameters.count || 10;
+
+  const handleOnChange = useCallback(
+    (evt: ResultsPaginationEvent): void => {
+      if (onChange) {
+        onChange(evt);
+      }
+      const { page, pageSize } = evt;
+      const offset = (page - 1) * pageSize;
+      setCurrentPage(page);
+      performSearch(
+        {
+          ...searchParameters,
+          count: pageSize,
+          offset
+        },
+        false
+      );
+    },
+    [onChange, performSearch, searchParameters]
+  );
+
+  // Since the Carbon pagination component controls the page number via its state
+  // without a way to influence it via props, we occasionally need to reset it back to "Page 1".
+  // This is done via a key on the CarbonPagination element, which is triggered via a counter
+  // that is incremented by this function.
+  const reset = useCallback(
+    ({ triggerOnChange }: ResetOptions) => {
+      setCurrentPage(1);
+      setResetCounter(resetCounter + 1);
+      if (triggerOnChange) {
+        handleOnChange({ page: 1, pageSize: resultsPerPage });
+      }
+    },
+    [handleOnChange, resetCounter, resultsPerPage]
+  );
+
+  // Externalize the reset function to a ref that the parent can send in,
+  // so that it can also reset the pagination as desired
+  useImperativeHandle(resetRef, () => ({ reset }));
 
   useEffect(() => {
     if (!!pageSize || !!resultsPerPage) {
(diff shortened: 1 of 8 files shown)


packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -126,17 +159,19 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
...
-            pageSize={actualPageSize}
+            pageSize={resultsPerPage}
             pageSizes={pageSizes}
             // NOTE: See PageSelector below for details about `onChange`
             onChange={handleOnChange}
             itemRangeText={handleItemRangeText}
             itemsPerPageText={mergedMessages.itemsPerPageText}
             pageRangeText={handlePageRangeText}
             pageText={handlePageText}
+            key={resetCounter}
```
noah-eigenfeld, line 174: Can you add a comment here or elsewhere explaining why you use a counter here?

noah-eigenfeld (approved):
Code changes look good, and some local testing showed this component working better

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -63,7 +82,49 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
...
+      if (triggerOnChange) {
+        handleOnChange({ page: 1, pageSize: resultsPerPage });
+      }
+    },
+    [handleOnChange, resetCounter, resultsPerPage]
+  );
+
+  // Externalize the reset function to a ref that the parent can send in,
+  // so that it can also reset the pagination as desired
+  useImperativeHandle(resetRef, () => ({ reset }));
```
jhpedemonte, line 127: There should be a mention of the available functions (just `reset` currently) that are available to call directly, in the JSDoc for this component.

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -34,6 +44,10 @@ export interface ResultsPaginationProps {
    * custom handler invoked when any input element changes in the ResultsPagination component
    */
   onChange?: (e: ResultsPaginationEvent) => void;
+  /**
+   * ref passed in by parent that can be used to reset Carbon's pagination component
```
jhpedemonte, line 48: Don't mention the carbon component; that's an implementation detail that shouldn't be exposed.

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -79,39 +140,13 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
     const pageFromOffset = Math.floor(actualOffset / actualPageSize) + 1;
     if (currentPage !== pageFromOffset) {
       setCurrentPage(pageFromOffset);
+      if (pageFromOffset === 1) {
+        reset({});
```
jhpedemonte, line 144: Why call reset here? If I go to page 2 then hit Prev to go back to page 1, this will call reset?

packages/discovery-react-components/src/components/ResultsPagination/__tests__/ResultsPagination.test.tsx
```diff
@@ -116,62 +116,6 @@ describe('ResultsPaginationComponent', () => {
         pageSize: 20
       });
     });
-
-    test('will add pageSize as a pageSize selection if it is not already included', async () => {
```
jhpedemonte, line 120: Why remove this ability? It seems odd that it was previously supported, but keeping it seems simple enough and prevents a breaking change.

packages/discovery-react-components/src/components/ResultsPagination/__tests__/ResultsPagination.test.tsx
```diff
@@ -116,62 +116,6 @@ describe('ResultsPaginationComponent', () => {
...
-        }
-      );
-
-      expect(getByText('25')).toBeInTheDocument();
-    });
-  });
-
-  describe('when there are component settings available', () => {
-    describe('and there are no display parameters passed on ResultsPagination', () => {
-      test('will update the count search param', async () => {
```
jhpedemonte, line 137: Why was this and following test removed?

ColbyJohnIBM: These are both for testing whether adjusting the `results_per_page` settings properly update the count value on the API calls. With this PR, we will no longer use `results_per_page` at all, in favor of using the `count` values from the shim, so these tests aren't applicable anymore.

packages/discovery-react-components/src/components/ResultsPagination/__tests__/ResultsPagination.test.tsx
ColbyJohnIBM: We needed this due to the mismatch of `results_per_page` and `count` values being sent in from the shim, but no longer need it. This functionality also caused some strange behaviors. For example, if the user adjusts the page size from a value that doesn't exist in the default list to one that does exist, the value is then removed from the list, which is strange behavior from the user's perspective (i.e. the dropdown menu options seem to change arbitrarily).

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
ColbyJohnIBM: I added a comment within the reset function.

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
ColbyJohnIBM: The `handleOnChange` function adjusts the `currentPage` and `offset` values at the same time, so `currentPage !== pageFromOffset` should never be true if the page change is being made via the pagination component itself. I'll double check that this portion of the code is not being hit in that case.

This piece is mostly to handle cases where something else is causing the offset to change outside of the user interacting with the pagination component (e.g. selecting a facet, creating a new query, etc.).

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -9,6 +19,12 @@ import onErrorCallback from 'utils/onErrorCallback';
 import { defaultMessages, Messages } from './messages';
 import { formatMessage } from 'utils/formatMessage';
 
+/**
+ * A pagination component to allow users to navigate through multiple pages of results
+ *
+ * Returns a reset function via the `resetRef` prop that resets the component to page 1.
```
jhpedemonte, line 25: This isn't true. It returns a component, but we externalize an API through use of `resetRef`.

Also, now that I think of it, this ref shouldn't be called `resetRef` since we could conceivably add other external API methods to that. Also, it leads to usage like:

```
<ResultsPagination resetRef={resetRef} ... />

...

resetRef.current.reset();
```

Double "reset"!

I don't think we need to be fancy here -- we can just use `ref`. Also, that'll clean up the code at the end of the file to just be:

```
export default forwardRef(ResultsPagination);
```

(But with TS)

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -34,6 +50,10 @@ export interface ResultsPaginationProps {
    * custom handler invoked when any input element changes in the ResultsPagination component
    */
   onChange?: (e: ResultsPaginationEvent) => void;
+  /**
+   * ref passed in by parent that can be used to reset the pagination component back to page 1
+   */
+  resetRef?: ForwardedRef<any>;
```
jhpedemonte, line 56:
```suggestion
  /**
   * Reference to imperative API
   */
  ref?: ForwardedRef<ResultsPaginationAPI>;
```

And define the type `ResultsPaginationAPI`. Oh, and export it from components (since it'll be used by consumers).

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -126,17 +167,19 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
       <>
         {!isResultsPaginationComponentHidden && (
           <CarbonPagination
-            className={classNames.join(' ')}
-            initialPage={currentPage}
```
jhpedemonte, line 130: Why remove `initialPage`?

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
```diff
@@ -63,7 +88,49 @@ const ResultsPagination: FC<ResultsPaginationProps> = ({
...
+    },
+    [onChange, performSearch, searchParameters]
+  );
+
+  // Since the Carbon pagination component controls the page number via its state
+  // without a way to influence it via props, we occasionally need to reset it back to "Page 1".
+  // This is done via a key on the CarbonPagination element, which is triggered via a counter
+  // that is incremented by this function.
+  const reset = useCallback(
+    ({ triggerOnChange }: ResetOptions) => {
```
jhpedemonte, line 119: Will a consumer of this component ever want to call `reset()` with `triggerOnChange: true`? Or is this just for internal usage only? If the latter, then don't expose it as part of the external API.

ColbyJohnIBM: Yes, we will need to use it in `discovery-tooling`. :+1:

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
ColbyJohnIBM: We weren't using it right-- it's a single-use prop that [sets the initial value of a state variable](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Pagination/experimental/Pagination.js#L39) in Carbon's experimental Pagination component. It's for cases where you want to load pagination on a page other than the first (i.e. the value defaults to 1). This was a key reason why our pagination wasn't working-- I think we assumed that this prop would set the page number as we changed the value, but it simply doesn't.

packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
ColbyJohnIBM: Agree that we can just go with `ref`. The end piece still needs to be a bit complex due to using an HOC and needing refs to be passed in.

ColbyJohnIBM: Unfortunately, this _did_ have to renamed to something besides `ref`: https://dev.to/justincy/using-react-forwardref-and-an-hoc-on-the-same-component-455m

I'll create a PR to revert this-- you can let me know what name suits you. 😄

PR #341, opened 2022-04-27
fix: add document id to virtual scroll key

noah-eigenfeld:
Contributes to https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/9712.

Fixes an issue where the naming of keys does not change when switching documents. This was causing styling issues, as the size of components would not be updated until the size of the window was adjusted.

Will require an update to the tooling repo, since this adds a new prop to the CIDocumentContent component (`documentId`).

I updated an existing storybook story based on the example in the connected issue (see above). Running the story for CIDocumentContent, do the following:
1. Change the knob for Document to ShortDocument. See that the styling looks normal.
2. Change the knob for Document to LongDocument. See that the segment holding the text looks like this:

  Screen Shot 2022-04-27 at 9 53 08 AM

  and not this:

  Screen Shot 2022-04-27 at 9 53 54 AM


packages/discovery-react-components/src/components/CIDocument/components/CIDocument/messages.ts
```diff
@@ -16,6 +17,7 @@ export interface Messages {
 
 export const defaultMessages: Messages = {
   defaultDocumentName: 'Document',
+  defaultDocumentId: 'Document Id',
```
ColbyJohnIBM, line 20: Is this a valid document id? What happens if this is left as the default?

ColbyJohnIBM (approved):

packages/discovery-react-components/src/components/CIDocument/components/CIDocument/messages.ts
noah-eigenfeld: >Is this a valid document id?

Not sure, but it only gets passed into the key value of the virtual scroll elements.

>What happens if this is left as the default?

Then this bug can still occur when the user is switching documents. But that would happen with any default value or just leaving it as `''`. Maybe it's better to just not have this and set the default inside the component. I was just following the existing pattern when I set this up, and didn't give it incredibly close consideration.

PR #343, opened 2022-05-02
fix: display DocumentPreview properly for documents with tables

ColbyJohnIBM:

- Show original document if highlighting table (don't need mappings)
- Add `data-child-begin` and `data-child-end` to table `article` tags
- Choose begin / end nodes as deepest part of DOM that contains offsets
- Update story to test file types with different highlight types

Run storybook and / or connect to tooling and verify that documents with tables are able to display in the DocumentPreview.

No, I think all changes are fixes without changing other significant behavior.


packages/discovery-react-components/src/utils/document/processDoc.ts
```diff
@@ -329,6 +329,10 @@ function setupSectionParser(
 
           if (tagName === TABLE_TAG) {
             sectionHtml.push('</article>');
+            sectionHtml[0] = sectionHtml[0].replace(
```
jhpedemonte, line 332: This concerns me. Are we sure it will always be index 0?

What about making use of `openTagIndices` instead, since that's already working? In `onopentag`, when `tagName === TABLE_TAG`, push to `openTagIndices`. Then later in `onclosetag`, do a pop and update, again in the `tagName === TABLE_TAG` section (which is then followed by an existing pop and updated).

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -95,20 +106,32 @@ export function getTextNodeAndOffset(node: Node, strOffset: number): NodeOffset
     domOffset = Math.max(0, domOffset - adjustment);
   }
 
+  console.log(
```
jhpedemonte, line 109: Remove this and the following console.log

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -95,20 +106,32 @@ export function getTextNodeAndOffset(node: Node, strOffset: number): NodeOffset
...
+    prevNode: Text | null;
   do {
+    // @ts-ignore
+    console.log(`Iterate through text nodes: len: ${len}, domOffset: ${domOffset}`);
+    prevNode = textNode;
     textNode = iterator.nextNode() as Text;
   } while (textNode && (len = textNode.data.length) && domOffset > len && (domOffset -= len));
 
-  if (textNode === null) {
+  // We didn't find any text nodes
```
jhpedemonte, line 127: Move this comment inside of the first `if` block; otherwise, it's ambiguous if it only applies to the first block or the entire if/else-if/else block.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -95,20 +106,32 @@ export function getTextNodeAndOffset(node: Node, strOffset: number): NodeOffset
...
+  if (textNode === null && prevNode === null) {
     throw new Error(`Failed to find text node. Node: ${node.textContent}, offset: ${strOffset}`);
+  } else if (textNode === null && prevNode !== null) {
+    return { textNode: prevNode, textOffset: prevNode.data.length };
```
jhpedemonte, line 131: Add a comment to this else-if about what is going on here, since it's not evident.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -28,27 +28,38 @@ export function findOffsetInDOM(
...
-    idx = 0;
-
-  do {
-    beginNode = allNodes[idx];
-  } while (
-    ++idx &&
-    idx < allNodes.length &&
-    begin >= parseInt(allNodes[idx].dataset.childBegin || '0', 10)
-  );
+  let beginNode = null,
```
jhpedemonte, line 31: Can you provide more details about why this was changed?

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -28,27 +28,38 @@ export function findOffsetInDOM(
...
+    endNode = null,
+    beginNodeSize = Number.MAX_VALUE,
+    endNodeSize = Number.MAX_VALUE;
+  allNodes.forEach((node: HTMLElement) => {
```
jhpedemonte, line 35: This will run through all of the nodes every time, even if the matches were at the beginning of the array. This should break out of the loop once we've found the correct matches.

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -42,16 +43,21 @@ export function isJsonFile(doc: QueryResult | null | undefined): boolean {
 /**
  * Returns the preview type for document
  */
-export function detectPreviewType(document: DiscoveryDocument, file?: DocumentFile): PreviewType {
+export function detectPreviewType(
+  document: DiscoveryDocument,
+  highlight: QueryResultPassage | QueryTableResult | undefined,
```
jhpedemonte, line 48: Instead of giving this a possible type of `undefined`, just make it an optional type:
```suggestion
  highlight?: QueryResultPassage | QueryTableResult,
```

That way, you don't need to add a 2nd `undefined` param to all of the tests, etc.

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -42,16 +43,21 @@ export function isJsonFile(doc: QueryResult | null | undefined): boolean {
...
   const hasPassage = !!document.document_passages?.[0]?.passage_text;
   const hasTextMappings = !!document.extracted_metadata?.text_mappings;
+  const isHighlightTable = isTable(highlight);
 
   // only render PDF if we have text mappings data
   if (fileType === 'pdf' && file) {
     // when hasTextMappings is true, that means the custom SDU model or OOB (CI) model is enabled
     // otherwise, that means the fast path
-    if (hasTextMappings || !hasPassage) {
+    if (hasTextMappings || !hasPassage || isHighlightTable) {
```
jhpedemonte, line 60: Update the comment above to take into consideration the 3 different possible options for rendering PDF.

packages/discovery-react-components/src/utils/document/documentUtils.ts
ColbyJohnIBM: We won't know that we've hit the smallest matches (i.e. the nodes lowest down the DOM tree) until we've checked every node, so we do need to check all the nodes. We could chop off the first chunk of nodes that have too small of `begin` values (using a binary search since the nodes are inherently sorted by `begin`), but I wasn't sure if the performance gains would be enough to be worth the code complexity. (The procedure would still be O(n), but this optimization would likely make it more efficient for larger documents and / or small highlights.)

packages/discovery-react-components/src/utils/document/documentUtils.ts
ColbyJohnIBM: Oh, yeah, sorry, accidentally left in from debugging a test issue.

jhpedemonte:
Got the following error when displaying a web crawled doc (with only default Doc Retrieval enrichments set):

Screen Shot 2022-05-03 at 11 23 50 AM

To reproduce:
- Go to `javped web crawl` project on STG
- Do an empty search
- Open first match. Renders as HTML, then shows error.

---

There are two issues here:
1. The fact that we are getting this error
2. The fact that this error brings down the entire UI

To fix (2), we either need to ensure that `findOffsetInDOM()` is always wrapped in a try/catch (or one of its ancestors), or we remove the `throw Error` entirely.

I believe in the past, we would sometimes get such errors but they were simply printed to the console -- the UI continued to function.

jhpedemonte:
I think the issue that the passage that is being shown and we want to highlight in the document is the title:

Screen Shot 2022-05-03 at 11 39 34 AM

But the title isn't rendered when displaying the document HTML. Not sure what to do about that.

jhpedemonte:
Using project `javped crawl + tables` (which has the web crawl of yarnpkg.com plus enabling Table Understanding enrichment), I verified that clicking `View table in document` rendered the doc in HTML and highlighted the table.

jhpedemonte:
One table highlight seems to be off. Empty search in `javped crawl + tables`, clicked on this one:

Screen Shot 2022-05-03 at 11 45 34 AM

The highlight of the table starts early (in the previous table):

Screen Shot 2022-05-03 at 11 46 35 AM

packages/discovery-react-components/src/utils/document/documentUtils.ts
ColbyJohnIBM: I ended up doing this differently, as even what I discussed above had at least one corner case that it would have missed. The new approach is more of a tree traversal with the help of a binary search to tell us where to "drill down" to look for each descendant. With the optimizations, this procedure tends to take <1ms to complete (if I'm not mistaken, it used to take on the order of ~5ms). Let me know if it's not clear from the comments how that procedure works.

ColbyJohnIBM:
> Got the following error when displaying a web crawled doc (with only default Doc Retrieval enrichments set):

I can no longer reproduce this with the latest change sin this PR.

> I think the issue that the passage that is being shown and we want to highlight in the document is the title. But the title isn't rendered when displaying the document HTML.

This regression existed in master and was fixed elsewhere.

> One table highlight seems to be off. The highlight of the table starts early (in the previous table).

I'm still seeing this issue on one of the table results. I'm looking into why it's occurring.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -208,6 +204,86 @@ export function uniqRects(rects: DOMRectList): Partial<DOMRectList> {
...
+      // Make sure that all nodes have begin / end values
+      nodes = nodes.filter(
+        node => node.dataset.childBegin !== undefined && node.dataset.childEnd !== undefined
+      );
+    }
+
+    // Perform binary search for the offset value on the array of nodes
+    // (This array of nodes will always be sorted by both
+    //  begin AND end, since sibling nodes are disjoint.)
+    const foundIndex = indexOfContainingNode(nodes, offset);
```
jhpedemonte, line 234: Can your custom binary-search code be replaced with one of the Lodash methods? https://lodash.com/docs/4.17.15#sortedIndex

jhpedemonte: If keeping your custom code, just have it return the node itself or null, instead of just the index.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -208,6 +204,86 @@ export function uniqRects(rects: DOMRectList): Partial<DOMRectList> {
...
+  // Set the low and high markers to the outer edges of the array
+  let lowIdx = 0,
+    highIdx = sortedArray.length - 1,
+    midIdx;
+
+  // Keep looping until the low marker passes the high marker
+  while (lowIdx <= highIdx) {
+    // Set the pivot to the halfway point of the outer markers
+    midIdx = Math.floor((highIdx + lowIdx) / 2);
+    if (parseInt(sortedArray[midIdx].dataset.childEnd || '0', 10) < offset) {
```
jhpedemonte, line 268: Refactor the childBegin/childEnd accessor into a separate function:

```js
function childBegin(node) {
  return parseInt(node.dataset.childBegin || '0', 10);
}
```

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -49,24 +50,27 @@ export function detectPreviewType(
...
   const isJsonType = isJsonFile(document);
   const isCsvType = isCsvFile(document);
   if (document.html && !isJsonType && !isCsvType) {
-    // HTML view cannot display a passage highlight unless the document have text_mappings.
-    // So, do not show as HTML when the document have a passage but does not have text_mappings.
-    if (hasTextMappings || !hasPassage) {
+    // When trying to highlight a passage, only show as HTML when the document has text_mappings.
+    // (since HTML view cannot display a passage highlight unless the document have text_mappings)
+    // When trying to highlight a table, text mappings aren't needed, so display HTML
+    if (!hasHighlight || (isHighlightPassage && hasTextMappings) || isHighlightTable) {
```
noah-eigenfeld, line 73: Maybe I'm confused, but won't the case where `!hasHighlight` always get swept up by the block of logic returning `PDF` at lines 62-64 (at time of writing). Meaning that case would never get caught here to return `HTML`.

packages/discovery-react-components/src/utils/document/documentUtils.ts
```diff
@@ -208,6 +204,86 @@ export function uniqRects(rects: DOMRectList): Partial<DOMRectList> {
...
 }
 
+/**
+ * Finds the descendant node that contains the given offset.
+ * This is done efficiently via a binary search.
+ * @param parentNode node to search the descendants of
+ * @param offset value to find within the nodes
+ * @return the descendant node, or null if none is found
+ */
+function findContainingNodeWithin(parentNode: HTMLElement, offset: number) {
```
noah-eigenfeld, line 214: Just a note: it might be nice to create some basic unit tests for these utils. Not required -- not sure how time consuming writing test cases would be for these.

noah-eigenfeld (approved):
Just left a few comments/questions. I don't think I found any deal-breakers, and the storybook stories seem to be working as expected.

packages/discovery-react-components/src/components/DocumentPreview/components/Highlight/typeUtils.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
ColbyJohnIBM: Hmmm, I think this was a copy-paste typo. I'll delete and check if things still work. :+1:

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
ColbyJohnIBM: Both of these if statements are nested within other ifs. For clarity, here's the code without the comments and git +/-.

```
if (fileType === 'pdf' && file) {
    if (!hasHighlight || hasTextMappings) {
      return 'PDF';
    }
}

if (document.html && !isJsonType && !isCsvType) {
  if (!hasHighlight || (isHighlightPassage && hasTextMappings) || isHighlightTable) {
    return 'HTML';
  }
}
```

So, a non-PDF, non-JSON, non-CSV file without a highlight and with html data will end up returning `'HTML'`. I considered condensing them to single if statements, but I thought that would be less clear (i.e. too many variables in one if).

packages/discovery-react-components/src/utils/document/documentUtils.ts
ColbyJohnIBM: Yeah, that probably makes sense. :+1:

packages/discovery-react-components/src/components/DocumentPreview/components/Highlight/typeUtils.ts
```diff
@@ -0,0 +1,10 @@
...
+import { QueryResultPassage, QueryTableResult } from 'ibm-watson/discovery/v2';
+
+export function isPassage(obj: any): obj is QueryResultPassage {
+  return obj && 'passage_text' in obj;
+}
+
+export function isTable(highlight: any): highlight is QueryTableResult {
+  return !!highlight?.table?.location;
+}
```
jhpedemonte, line 10: Were these moved to prevent circular-dependency? If so, you may now want to check other parts of the code that do this code and replace with calls to `isPassage` or `isTable`.

packages/discovery-react-components/src/components/DocumentPreview/utils/documentData.ts
```diff
@@ -49,24 +50,27 @@ export function detectPreviewType(
 ): PreviewType {
   const fileType = document.extracted_metadata?.file_type;
   // passages contain location offsets against text-based strings (not HTML)
-  const hasPassage = highlight && 'passage_text' in highlight;
   const hasTextMappings = !!document.extracted_metadata?.text_mappings;
+  const hasHighlight = !!highlight;
+  const isHighlightPassage = isPassage(highlight);
+  const isHighlightTable = isTable(highlight);
```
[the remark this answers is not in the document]
ColbyJohnIBM: Sure, that's fine. :+1:

packages/discovery-react-components/src/components/DocumentPreview/components/Highlight/typeUtils.ts
ColbyJohnIBM: Yes, that was the reason. I looked for uses of the old functions, but I'll also look for code that seems to be checking these same values. :+1:

jhpedemonte (approved):
Beyond things noted, code looks good and functions as expected.

PR #344, opened 2022-05-03
fix: render correct doc to show passage highlights

jhpedemonte:

Render TEXT when given an HTML file with passage to highlight (which only works on text, when there is no `text_mappings` to map to HTML-based offsets).

Run example app against a collection of web-crawled HTML files, where SDU model is default text-only.

---

#### To do:
- [ ] Update tests
- [ ] Add test to make sure we don't regress this in the future.


noah-eigenfeld:
Screen Shot 2022-05-04 at 2 30 45 PM
Tested locally by linking with tooling. Seems to be fixed.

PR #351, opened 2022-05-24
fix: add 'icp4d-api' path for auth url

jhpedemonte:

CP4D now requires the base path `/icp4d-api` for the auth url

1. Run `./runExampleApp.sh` against a CP4D deployment
2. Follow the "Manual setup" instructions against a CP4D deployment

Yes: updated README

### DIFF
--- a/examples/discovery-search-app/scripts/setSdkUrl.js
+++ b/examples/discovery-search-app/scripts/setSdkUrl.js
@@ -1,7 +1,6 @@
 const fs = require('fs');
 const path = require('path');
 const axios = require('axios').default;
-const https = require('https');
 const urlJoin = require('proper-url-join');
 const { validateSdkEnvVars } = require('./validateSdkEnvVars.js');
 const ibmCredentialsFilePath = path.join(__dirname, '../', 'ibm-credentials.env');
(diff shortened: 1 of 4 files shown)


examples/discovery-search-app/scripts/setSdkUrl.js
```diff
@@ -29,50 +28,60 @@ function writeEnvVarToFileAndReturn(key, value) {
...
+
+    let baseUrl = authUrl;
+    if (authUrl.endsWith('/icp4d-api')) {
+      baseUrl = authUrl.substring(0, authUrl.length - '/icp4d-api'.length);
+    }
+
+    let globalRejectUnauthorized = process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+    if (!rejectUnauthorized) {
+      process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+    }
```
jhpedemonte, line 46: For some reason, the previous code of creating an HTTPS agent with option `rejectUnauthorized: false` does not work with `axios.post()`. Only solution I found was to set `NODE_TLS_REJECT_UNAUTHORIZED`.

runExampleApp.sh
```diff
@@ -4,8 +4,8 @@ set -e
 
 PREREQUISITES=("yarn" "node")
 SCRIPT_DIR=$(
-    cd "$(dirname "$0")"
-    pwd
+  cd "$(dirname "$0")"
```
jhpedemonte, line 7: Most changes in this file are linter/prettier. The only functional change is at line 110.

ColbyJohnIBM (approved):
Looks good, but LGTM has flagged something that needs fixing. With that fixed, approved.

examples/discovery-search-app/scripts/setSdkUrl.js
ColbyJohnIBM: LGTM doesn't like disabling certificate validation-- what about `rejectUnauthorized: false` doesn't work with `axios.post`?

PR #366, opened 2022-06-13
render textLayer optionally

agusev:

- Implements the following [issue](https://github.ibm.com/watson-discovery/disco-issue-tracker/issues/11135)
- allows disabling `textLayer` rendering
- use `Storybook` within the issue it implements to verify that `textLayer` is NOT rendered 
- use `Storybook` within the `discovery-components` to verify that `textLayer` is rendered by default

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
@@ -47,7 +47,7 @@ interface Props extends WithErrorBoundaryProps {
    */
   highlight?: QueryResultPassage | QueryTableResult;
   /**
-   * disable textLayer if true
+   * Disable the text layer overlay when rendering PDF (defaults to `false`)
    */
   disableTextLayer?: boolean;
   /**
--- a/packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
@@ -47,6 +47,10 @@ interface Props extends WithErrorBoundaryProps {
    */
   highlight?: QueryResultPassage | QueryTableResult;
   /**
+   * disable textLayer if true
+   */
+  disableTextLayer?: boolean;
+  /**
    * i18n messages for the component
    */
   messages?: Messages;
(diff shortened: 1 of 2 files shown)


ColbyJohnIBM (approved):

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -56,6 +61,7 @@ const PdfViewer: FC<Props> = ({
   scale,
   document,
   textLayerClassName,
+  disableTextLayer,
```
ColbyJohnIBM, line 64: I know that `undefined` is falsy, but it might make sense to give this a default value of `false`.

noah-eigenfeld (approved):

packages/discovery-react-components/src/components/DocumentPreview/DocumentPreview.tsx
```diff
@@ -46,6 +46,10 @@ interface Props extends WithErrorBoundaryProps {
    * `document.document_passages` or `searchResults.table_results`.
    */
   highlight?: QueryResultPassage | QueryTableResult;
+  /**
+   * disable textLayer if true
+   */
+  disableTextLayer?: boolean;
```
jhpedemonte, line 52: - Specify that the default is `false`
- This property is at the `DocumentPreview` level, which can render a document in several different forms. But this property only applies to `PdfViewer`. Given that, we should be more specific about its intent.
- Follow convention for these comments of using sentence case
 
So something like this:
```suggestion
  /**
   * Disable the text layer overlay when rendering PDF (defaults to `false`)
   */
  disablePdfTextLayer?: boolean;
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -31,6 +31,11 @@ type Props = PdfDisplayProps & {
    */
   textLayerClassName?: string;
 
+  /**
+   * disable textLayer if true
```
jhpedemonte, line 35: Similar comments to the component above (sentence case, default value). But fine to keep this named `disableTextLayer`, since adding `Pdf` in there would be redundant for `PdfViewer` component.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -112,16 +118,22 @@ const PdfViewer: FC<Props> = ({
         <canvas
           ref={canvasRef}
           className={`${base}__canvas`}
-          style={{ width: `${canvasInfo?.width ?? 0}px`, height: `${canvasInfo?.height ?? 0}px` }}
+          style={{
+            width: `${canvasInfo?.width ?? 0}px`,
+            height: `${canvasInfo?.height ?? 0}px`,
+            transform: `scale(${scale})`
```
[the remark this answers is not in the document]
agusev: `Transform` allows `canvas` to be scaled.

PR #379, opened 2022-07-16
fix: revert to renaming ref prop with forwardRef and HOC

ColbyJohnIBM:

- rename the `ref` prop back to `resetRef` to allow it to be passed in properly
(Fix bug from https://github.com/watson-developer-cloud/discovery-components/pull/340#discussion_r922628853)

Contributes to [Watson-Discovery/disco-issue-tracker#4145](https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/4145)

Link to local tooling, make sure the resetRef prop is actually  getting populated (doesn't remain `undefined`).

Shouldn't need additional documentation.

No.

### DIFF
--- a/packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
+++ b/packages/discovery-react-components/src/components/ResultsPagination/ResultsPagination.tsx
@@ -53,7 +53,7 @@ export interface ResultsPaginationProps {
   /**
    * Reference to imperative API
    */
-  resetRef?: ForwardedRef<ResultsPaginationAPI>;
+  apiRef?: ForwardedRef<ResultsPaginationAPI>;
   /**
    * Additional props to be passed into Carbon's Pagination component
    */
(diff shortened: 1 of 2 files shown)


jhpedemonte:
I'm confused, why would the name matter? (I still don't like this being named `resetRef`)

Documentation and other implementations seemed fine with just `ref`.

jhpedemonte:
What if you use the `ErrorBoundary` component instead of the HOC? https://github.com/bvaughn/react-error-boundary#usage

ColbyJohnIBM:
Doing something like this still requires a different name for ref. Again, not sure exactly why, but passing a ref through a few layers of components seems to cause issues if it's named `ref` everywhere.

```
const ResultsPaginationWithBoundary = (props: ResultsPaginationProps, ref: ForwardedRef<any>) => (
  <ErrorBoundary
    FallbackComponent={FallbackComponent('ResultsPagination')}
    onError={onErrorCallback}
  >
    <ResultsPagination {...props} resetRef={ref} />
  </ErrorBoundary>
);

export default forwardRef<any, ResultsPaginationProps>(ResultsPaginationWithBoundary);
```

jhpedemonte (approved):
One suggestion. With that, approved.

PR #385, opened 2022-08-01
fix: adjust how canvas dimensions are computed

ColbyJohnIBM:

- Set initial canvas size based on the available width
- Adjust the size of the text layer and highlights proportionally
- Add ResizeObserver to adjust canvas accordingly when page is resized

Contributes to https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/12758

- Run UI locally linked with components repo running with this branch.
- Check that document highlighting, underlining, and boxing all align with PDF documents, regardless of zooming in and out
- Check that Storybook looks correct and that zooming works correctly for all stories

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
@@ -1,8 +1,10 @@
-import React, { FC, useEffect, useRef, useMemo, useCallback } from 'react';
+import React, { FC, useEffect, useRef, useCallback, useMemo, useState, RefObject } from 'react';
 import cx from 'classnames';
 import PdfjsLib, { PDFDocumentProxy, PDFPageProxy, PDFPromise, PDFRenderTask } from 'pdfjs-dist';
 import PdfjsWorkerAsText from 'pdfjs-dist/build/pdf.worker.min.js';
 import { settings } from 'carbon-components';
+import debounce from 'lodash/debounce';
+import useResizeObserver from '@react-hook/resize-observer';
 import useAsyncFunctionCall from 'utils/useAsyncFunctionCall';
 import { QueryResult } from 'ibm-watson/discovery/v2';
 import { DocumentFile } from '../../types';
@@ -10,6 +12,8 @@ import { getTextMappings } from '../../utils/documentData';
 import PdfViewerTextLayer, { PdfRenderedText } from './PdfViewerTextLayer';
 import { toPDFSource } from './utils';
 import { PdfDisplayProps } from './types';
+
+const RESIZE_DEBOUNCE = 50;
 
 setupPdfjs();
 
@@ -69,6 +73,8 @@ const PdfViewer: FC<Props> = ({
   children
 }) => {
   const canvasRef = useRef<HTMLCanvasElement>(null);
+  const rootRef = useRef<HTMLDivElement>(null);
+  const [canvasInfo, setCanvasInfo] = useState<CanvasInfo | null>(null);
 
   const loadedFile = useAsyncFunctionCall(
     useCallback(async () => (file ? await _loadPdf(file) : null), [file])
@@ -80,7 +86,15 @@ const PdfViewer: FC<Props> = ({
     )
   );
 
-  const canvasInfo = useMemo(() => getCanvasInfo(loadedPage, scale), [loadedPage, scale]);
+  useEffect(
+    () => setCanvasInfo(getCanvasInfo(loadedPage, scale, rootRef)),
+    [loadedPage, scale, rootRef]
+  );
+
+  useResizeObserver(
+    rootRef.current,
+    debounce(() => setCanvasInfo(getCanvasInfo(loadedPage, scale, rootRef)), RESIZE_DEBOUNCE)
+  );
 
   // render page
   useAsyncFunctionCall(
@@ -111,17 +125,18 @@ const PdfViewer: FC<Props> = ({
     }
   }, [setHideToolbarControls]);
 
+  const proportion = canvasInfo?.proportion || 1;
+
   const base = `${settings.prefix}--document-preview-pdf-viewer`;
   return (
-    <div className={cx(base, className)}>
+    <div ref={rootRef} className={cx(base, className)}>
       <div className={`${base}__wrapper`}>
         <canvas
           ref={canvasRef}
           className={`${base}__canvas`}
           style={{
             width: `${canvasInfo?.width ?? 0}px`,
-            height: `${canvasInfo?.height ?? 0}px`,
-            transform: `scale(${scale})`
+            height: `${canvasInfo?.height ?? 0}px`
           }}
           width={canvasInfo?.canvasWidth}
           height={canvasInfo?.canvasHeight}
@@ -130,11 +145,11 @@ const PdfViewer: FC<Props> = ({
           <PdfViewerTextLayer
             className={cx(`${base}__text`, textLayerClassName)}
             loadedPage={loadedPage}
-            scale={scale}
+            scale={scale * proportion}
             setRenderedText={setRenderedText}
           />
         )}
-        {children}
+        {typeof children === 'function' ? children({ proportion }) : children}
       </div>
     </div>
   );
@@ -216,20 +231,24 @@ type CanvasInfo = {
   height: number;
   canvasWidth: number;
   canvasHeight: number;
+  proportion: number;
   viewport: PdfjsLib.PDFPageViewport;
 };
 
 function getCanvasInfo(
   loadedPage: PdfjsLib.PDFPageProxy | null | undefined,
-  scale: number
+  scale: number,
+  rootRef: RefObject<HTMLDivElement>
 ): CanvasInfo | null {
   const canvasScale = window.devicePixelRatio ?? 1;
   const viewport = loadedPage?.getViewport({ scale: scale * canvasScale });
-  if (viewport) {
+  const rootDimensions = rootRef?.current?.getBoundingClientRect();
+  if (viewport && rootDimensions) {
     const { width: canvasWidth, height: canvasHeight } = viewport;
-    const width = Math.ceil(canvasWidth / canvasScale);
-    const height = Math.ceil(canvasHeight / canvasScale);
-    return { width, height, canvasWidth, canvasHeight, viewport };
+    const width = rootDimensions.width * scale;
+    const height = (width * canvasHeight) / canvasWidth;
+    const proportion = (width / canvasWidth) * canvasScale;
+    return { width, height, canvasWidth, canvasHeight, proportion, viewport };
   }
   return null;
 }
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
@@ -1,10 +1,9 @@
-import React, { FC, useEffect, useRef, useCallback, useMemo, useState, RefObject } from 'react';
+import React, { FC, useEffect, useRef, useCallback, useMemo, useState } from 'react';
 import cx from 'classnames';
 import PdfjsLib, { PDFDocumentProxy, PDFPageProxy, PDFPromise, PDFRenderTask } from 'pdfjs-dist';
 import PdfjsWorkerAsText from 'pdfjs-dist/build/pdf.worker.min.js';
 import { settings } from 'carbon-components';
-import debounce from 'lodash/debounce';
-import useResizeObserver from '@react-hook/resize-observer';
+import useSize from '@react-hook/size';
 import useAsyncFunctionCall from 'utils/useAsyncFunctionCall';
 import { QueryResult } from 'ibm-watson/discovery/v2';
 import { DocumentFile } from '../../types';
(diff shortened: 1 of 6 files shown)


agusev (approved):
tested in `Storybook`. Works as expected

jhpedemonte:
**Functional:**

1. There seems to be some sort of timing issue: in storybook, in **DocumentPreview**, I switch between **default** and **passage highlighting**, giving each time to load. But every so often (fairly regularly), the render gets stuck on the loading skeleton. I cannot replicate this on the published storybook.
2. Text for rendered PDF now looks very blurry. I thought that Susumu had put in a fix for that, but I notice it with the current `master` code as well, when zooming in. However, this PR exacerbates things since the "default" rendering size will now not be the PDF's 100%, but whatever the parent container is. We should see if there's a way to fix this so it doesn't look so bad after this is merged.

ColbyJohnIBM:
I believe I've fixed both of the issues that you mentioned. I could previously reproduce both, but now cannot reproduce either. I believe the blurriness was caused by not adjusting the scale of the viewport accordingly with the width change and the loading issue was caused by unnecessary additional computations of the canvas info, which was causing additional renders.

Please take another look when you get a chance. :+1:

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -80,7 +87,21 @@ const PdfViewer: FC<Props> = ({
...
+  useResizeObserver(
+    rootRef.current,
+    debounce(() => {
+      const currentRootWidth = rootRef?.current?.getBoundingClientRect().width;
+      if (!!currentRootWidth && currentRootWidth !== previousRootWidth) {
+        setCanvasInfo(getCanvasInfo(loadedPage, scale, rootRef));
+        setPreviousRootWidth(currentRootWidth);
+      }
+    }, RESIZE_DEBOUNCE)
+  );
```
jhpedemonte, line 104: I noticed that the author of `useResizeObserver` also has a hook called `useSize` (which uses the former). I think that might be better here. Something like:

```js
const [width] = useSize(rootRef);
useEffect(() => {
  // width has changed; update canvas info
  setCanvasInfo(getCanvasInfo(loadedPage, scale, width)); // NOTE: just pass width, since we know it already
}, [width]);
```

jhpedemonte: Oh, this also takes care of the `useEffect` above this code (since it's the same!).

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +238,28 @@ type CanvasInfo = {
...
+  scale: number,
+  rootRef: RefObject<HTMLDivElement>
 ): CanvasInfo | null {
-  const canvasScale = window.devicePixelRatio ?? 1;
-  const viewport = loadedPage?.getViewport({ scale: scale * canvasScale });
-  if (viewport) {
+  const rootDimensions = rootRef?.current?.getBoundingClientRect();
+  if (loadedPage && rootDimensions) {
+    const canvasScale = window.devicePixelRatio ?? 1;
+    const width = rootDimensions.width * scale;
```
jhpedemonte, line 253: Let's pick some better names here:
```suggestion
    const scaledWidth = rootDimensions.width * scale;
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +238,28 @@ type CanvasInfo = {
...
+    const pageWidth = loadedPage.view[2];
```
jhpedemonte, line 254: What is `view[2]`?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +238,28 @@ type CanvasInfo = {
...
+    const proportion = rootDimensions.width / pageWidth;
+
+    const viewport = loadedPage.getViewport({
+      scale: (scale * canvasScale * width) / pageWidth
```
jhpedemonte, line 258: This is weird. `width` is set to `rootDimensions.width * scale`, so then this becomes:

```js
scale * canvasScale * rootDimensions.width * scale / pageWidth
```

So multiplying by `scale` twice?

Given what was there in the previous code, it seems like what you want to do is:

```js
scale * canvasScale * proportion
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +238,28 @@ type CanvasInfo = {
...
+    });
     const { width: canvasWidth, height: canvasHeight } = viewport;
-    const width = Math.ceil(canvasWidth / canvasScale);
```
jhpedemonte, line 230: So `width` is never based on `canvasWidth` any more?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +238,28 @@ type CanvasInfo = {
...
-    const height = Math.ceil(canvasHeight / canvasScale);
-    return { width, height, canvasWidth, canvasHeight, viewport };
+    const height = (width * canvasHeight) / canvasWidth;
```
jhpedemonte, line 261: Why is `height` calculated using width??

ColbyJohnIBM: The viewport gives us a height and width of the canvas. `canvasWidth / canvasHeight` gives us the height:width ratio for the specific document that we're looking at. Multiplying that by the display width gives us the display height that the document should have.

E.g. say we have a document that is 850px w x 1100px h (8.5" x 11" page), and we want to display it in a window of width 1000px. To determine the proper height, we would take 1100px (the PDF's actual height), divide by 850px (the PDF's actual width), then multiply by 1000px (the width of the space that we have to display it) = ~1294px for the display height.

In short, we need the width to compute the height because the size of the displayed PDF is entirely determined by the width now.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: Good question, and I'm glad you asked. I thought it was the inherent width of the PDF, but it's actually the `x2` of the visible portion of the PDF (https://mozilla.github.io/pdf.js/api/draft/module-pdfjsLib-PDFPageProxy.html#view). So we'll need to adjust this to `loadedPage.view[2] - loadedPage.view[0]`, and make sure there's no unit conversion.

The reason that this is necessary is that we need to determine the ratio of available width to the real width of the PDF in order to incorporate that value into the scale being sent into `getViewport`.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: Do you want the values returned by this function to be renamed accordingly as well (line 262)?

`return { width, height, canvasWidth, canvasHeight, proportion, viewport };`

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: Oh, nice. Didn't notice that, that could be very convenient. I'll check out how well it handles the page resizes (from a debouncing perspective)...

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: I'll have to dig a bit on this to make sure we're setting this properly to avoid any blurriness, but `scale * canvasScale * proportion` sounds right at the moment...

ColbyJohnIBM:
I'll make adjustments based on what you've pointed out and add some comments to make it more clear what each value is in the canvas info function.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
jhpedemonte: This concerns me. "Unit conversion", "is often", etc. Is there no function that returns what you need in a predictable manner?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
jhpedemonte: Ah right. Add a comment about that.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
jhpedemonte: Gotcha. Then clean up the code to make this evident. I would start by setting:

```js
const pageAspectRatio = canvasHeight / canvasWidth;
```

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
jhpedemonte: Well, just looking at that line, it is not evident to me how `width` and `canvasWidth` are different or what they represent. Also, what does `proportion` stand for?

Choosing more specific names may help, or maybe we can only handle these definitions through comments. Or both!

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
jhpedemonte: If you still need to debounce, you can simply move that to around `setCanvasInfo`.

That author also has a `useDebounce` hook, which may be useful.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: We could create one viewport with a scale value of `scale * canvasScale` and use its width / height to generate the `proportion` value necessary to create a second viewport with scale value of `scale * canvasScale * proportion`, but that seemed convoluted.

As far as I can tell, the unit conversion is what we're already doing with the device pixel ratio. `loadedPage.view` will often be `[0, 0, width, height]`, since the PDF will usually start at (0,0) and then continue to (width, height). But occasionally, it would go from (x1, y1) to (x1 + width, y1 + height) `loadedPage.view = [x1, y1, x1 + width, y1 + height]`. Setting `pageWidth = loadedPage.view[2] - loadedPage.view[0]` should be correct in either case, even if we're often hitting the first case (in which `pageWidth = loadedPage.view[2]` would be sufficient.

ColbyJohnIBM: To be clear, we need the `proportion` value prior to the creation of the real viewport to prevent the blurriness that you were seeing previously.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: Yeah, I mentioned to you that I'd need help naming `proportion` something more descriptive. I'll go ahead and push my changes that include comments and see if things a re more clear...

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
ColbyJohnIBM: I think the debounce may not be absolutely necessary-- resizing the window doesn't cause the PDF to jump around too much. We can wait to see if it bothers anyone (unlikely).

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +224,47 @@ type CanvasInfo = {
...
   loadedPage: PdfjsLib.PDFPageProxy | null | undefined,
-  scale: number
+  scale: number,
+  rootWidth: number
 ): CanvasInfo | null {
-  const canvasScale = window.devicePixelRatio ?? 1;
-  const viewport = loadedPage?.getViewport({ scale: scale * canvasScale });
-  if (viewport) {
+  if (loadedPage) {
+    // Set the displayed width of the canvas to fit the available width when scale = 1,
```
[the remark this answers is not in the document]
ColbyJohnIBM: I was wondering how to express this properly. It essentially fills the space that the parent of `PDFViewer` gives it. What's a good way to express that concisely?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -216,20 +224,47 @@ type CanvasInfo = {
...
+    // and to scale up or down accordingly with the scale value
+    const width = rootWidth * scale;
+
+    // The "real" width of the PDF page can be found by subtracting the x2 and x1 values
```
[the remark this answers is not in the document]
ColbyJohnIBM: I sat for a while trying to find a good word for that ("raw" kept getting stuck in my head, but didn't feel right). "Native" is a great option.

PR #397, opened 2022-09-12
feat: allow displaying multiple passages from each document

ColbyJohnIBM:

- refactor Result component to handle multiple passages
- stack passage and table results vertically
- adjust styling accordingly
- update tests / Storybook

- View search components in Storybook and make sure they look and behave correctly. Note that the SearchResults story now has a result with two passages.
- Link to the corresponding `discovery-tooling` PR and make sure search results look correct when following the instructions in that PR.


examples/discovery-search-app/cypress/fixtures/query/passageResults.json
```diff
@@ -24,19 +24,19 @@
         "file_type": "pdf"
       },
       "text": [
-        "This is a document.\nThis result multiple passages, but you should only be able to see the first one. IBM if you can see this passage, something probably borked"
+        "This is a document.\nThis result multiple passages, so we'll display them in the order that they're returned. IBM We now display all passages that are returned for each document."
```
jhpedemonte, line 27: These sentences are hurting me.
```suggestion
        "This is a document.\nThis result has multiple passages, so we'll display them in the order that they're returned. IBM. We can display multiple passages that are returned for each document."
```

Also removed "all" since it depends on the setting.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -102,26 +101,32 @@ export const Result: React.FunctionComponent<ResultProps> = ({
...
-    'document_passages[0]'
+    'document_passages'
+  )?.filter(passage => !!get(passage, 'passage_text'));
+  const passageTexts: (string | undefined)[] | undefined = passages?.map(passage =>
+    get(passage, 'passage_text')
   );
-  const firstPassageText = get(firstPassage, 'passage_text');
-  const hasPassage = usePassages && !!firstPassageText;
-  let displayedText: string | undefined = get(result, bodyField);
+  const hasPassage = usePassages && !!passageTexts && passageTexts.length > 0;
```
noah-eigenfeld, line 111: Nit, since you're already updating the singular vars to plural
```suggestion
  const hasPassages = usePassages && !!passageTexts && passageTexts.length > 0;
```

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -102,26 +101,32 @@ export const Result: React.FunctionComponent<ResultProps> = ({
...
+  let displayedTexts: string[] | undefined = get(result, bodyField)
+    ? [get(result, bodyField)]
+    : undefined;
+  let displayedTextElements: SelectedResult['element'][] | null = null;
+  let displayedTextElementType: SelectedResult['elementType'] = null;
   if (hasPassage) {
```
noah-eigenfeld, line 117: If you decide to use the above nit
```suggestion
  if (hasPassages) {
```

noah-eigenfeld (approved):
Tested locally, and functionality looks good. Left a few nit suggestions, since you still have a build failure to sort out

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -102,26 +101,32 @@ export const Result: React.FunctionComponent<ResultProps> = ({
     fetchDocumentsResponseStore: { isLoading }
   } = useContext(SearchContext);
 
-  const firstPassage: DiscoveryV2.QueryResultPassage | undefined = get(
+  const passages: DiscoveryV2.QueryResultPassage[] | undefined = get(
     result,
-    'document_passages[0]'
+    'document_passages'
+  )?.filter(passage => !!get(passage, 'passage_text'));
+  const passageTexts: (string | undefined)[] | undefined = passages?.map(passage =>
```
jhpedemonte, line 108: So this can be an array of `undefined`? That doesn't seem right.

jhpedemonte (approved):
Works well and code generally looks good. Have some small suggestions, though. With those fixed, approved.

jhpedemonte:
FYI: I opened the story **SearchResults > default** and opened the "canvas" in it's own tab, then ran a11y check. If found an issue with the table display:

> The <table> element with "presentation" role or "none" role has structural element(s) and/or attribute(s) 'th'

Also had issues with the "View" links:

> Content is not within a landmark element

examples/discovery-search-app/cypress/integration/passages/spec.ts
```diff
@@ -13,9 +13,11 @@ describe('Passage Results', () => {
 
     // SearchResults displays ONLY the first passage text of the results that have passages
     cy.findByText(
-      'This result multiple passages, but you should only be able to see the first one.'
+      "This result multiple passages, so we'll display them in the order that they're returned."
```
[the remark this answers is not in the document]
ColbyJohnIBM: Generally, our linter is set up to prefer single quotes, but if there's a single quote in the string itself, the convention is to use double quotes rather than escaping with `/'`.

From the prettier docs:
> If the number of quotes outweighs the other quote, the quote which is less used will be used to format the string - Example: "I'm double quoted" results in "I'm double quoted" and "This \"example\" is single quoted" results in 'This "example" is single quoted'.

https://prettier.io/docs/en/options.html#quotes

packages/discovery-react-components/src/components/SearchResults/__fixtures__/searchResults.ts
```diff
@@ -34,7 +34,11 @@ const searchResults: DiscoveryV2.QueryResponse = {
       document_passages: [
         {
           passage_text:
-            "Deep learning is a specific method of the <em>machine learning</em> that incorporates nueral networks in successive layers from data in an iterative manner. Deep learning is especially userful when you're tring to learn patterns from unstructured data"
+            'Deep learning is a specific method of <em>machine learning</em> that incorporates neural networks in successive layers from data in an iterative manner.'
+        },
+        {
+          passage_text:
+            "Deep learning is especially useful when you're trying to learn patterns from unstructured data."
```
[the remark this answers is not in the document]
ColbyJohnIBM: Same explanation as above.

examples/discovery-search-app/cypress/fixtures/query/passageResults.json
ColbyJohnIBM: I'm fine with the language change, but the component will actually display all of the passages returned (but the number returned will be limited by `max_per_document`).

examples/discovery-search-app/cypress/fixtures/query/passageResults.json
```diff
@@ -24,19 +24,19 @@
...
       ],
       "document_passages": [
         {
-          "passage_text": "This result multiple passages, but you should only be able to see the first one. <em>IBM</em>",
+          "passage_text": "This result multiple passages, so we'll display them in the order that they're returned. <em>IBM</em>",
```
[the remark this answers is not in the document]
ColbyJohnIBM: Haha, my sentence without a verb is clearly superior.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -102,26 +101,32 @@ export const Result: React.FunctionComponent<ResultProps> = ({
...
+  const hasPassage = usePassages && !!passageTexts && passageTexts.length > 0;
+  let displayedTexts: string[] | undefined = get(result, bodyField)
+    ? [get(result, bodyField)]
+    : undefined;
+  let displayedTextElements: SelectedResult['element'][] | null = null;
+  let displayedTextElementType: SelectedResult['elementType'] = null;
   if (hasPassage) {
-    displayedText = firstPassageText;
+    displayedTexts = passageTexts as string[];
+    displayedTextElements = passages as DiscoveryV2.QueryResultPassage[];
```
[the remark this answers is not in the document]
ColbyJohnIBM: Yes, that was the reason for `as`-- I forgot about that use of `!`. 😬

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
ColbyJohnIBM: This is because `get` could technically return `undefined`. (Even with optional chaining, TS will think that this could return `undefined`.) However, we know that it can't since we already filtered out that possibility on line 107.

ColbyJohnIBM: I tried to build out an interface that extended `DiscoveryV2.QueryResultPassage`, but required `passage_text`, but hit a snag (forgetting exactly what). I'll try that again and see how it goes this time...

examples/discovery-search-app/cypress/fixtures/query/passageResults.json
jhpedemonte: Right, that's what I was getting it: it will be limited by `max_per_document`, so not technically "all".

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
ColbyJohnIBM: `!` operator is useful here as well. :+1:

ColbyJohnIBM:
The a11y issues already exist [on the hosted Storybook](https://watson-developer-cloud.github.io/discovery-components/storybook/iframe.html?id=searchresults--default), so are not introduced by these changes. We should create an issue for a11y issues in the components repo / stories and address these there.

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
```diff
@@ -83,7 +83,7 @@ export class SearchFilterTransform {
   private static unquoteString(quotedString: string): string {
     return quotedString
       .replace(new RegExp(SearchFilterTransform.STRING_IN_QUOTES), '$1')
-      .replace(/\\"/, '"');
+      .replace(/\\"/g, '"');
```
jhpedemonte, line 86: What's up with these changes? They make sense, just wondering how we didn't catch them before.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -102,26 +102,28 @@ export const Result: React.FunctionComponent<ResultProps> = ({
...
-  let displayedText: string | undefined = get(result, bodyField);
-  if (hasPassage) {
-    displayedText = firstPassageText;
+  const passages: DiscoveryV2.QueryResultPassage[] =
+    result?.document_passages?.filter(passage => !!passage?.passage_text) || [];
+  const passageTexts: string[] = passages?.map(passage => passage.passage_text!) || [];
+  const hasPassages = usePassages && !!passageTexts && passageTexts.length > 0;
+  let displayedTexts: string[] | undefined = get(result, bodyField)
+    ? [get(result, bodyField)]
+    : undefined;
```
jhpedemonte, line 111: You can still replace this `get()` with optional chaining:
```suggestion
  let displayedTexts: string[] | undefined = result?.[bodyField]
    ? [result?.[bodyField]]
    : undefined;
```

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
```diff
@@ -178,18 +178,24 @@ export const Result: React.FunctionComponent<ResultProps> = ({
...
-                dangerouslyRenderHtml={shouldDangerouslyRenderHtml}
-              />
-            )}
+            {displayedTexts &&
+              !showTablesOnlyResults &&
+              displayedTexts.map((displayedText, index) => {
+                const displayedTextElement = displayedTextElements?.[index];
+                return (
+                  <ResultElement
+                    key={md5(displayedText)}
```
jhpedemonte, line 187: It's conceivable that there could be 2+ passages with the same text, which would break this. Using offsets would help (since those are unique) -- but you mentioned to me (elsewhere) that it was difficult to implement.

jhpedemonte (approved):

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
ColbyJohnIBM: I was hitting issues where the values being sent in were of the form `abc.def`-- and I wasn't sure how to handle that with optional chaining. I.e. `result?.[bodyField]` was coming back `undefined` in those cases (since it was looking for a single field called `abc.def`, rather than two layers of fields).

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
ColbyJohnIBM: Oops, adding this change to this PR was actually accidental-- I had made it when investigating the issue Fucheng had pointed out last week. We can leave it here or wait until we're fixing that issue. (This change alone doesn't fix it.)

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
jhpedemonte: Ah, that's right. Forgot that `bodyField` can have multiple levels. Ignore my suggestion, then.

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
jhpedemonte: If not specifically for this (and you'll handle it later), I would prefer to remove it.

packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
ColbyJohnIBM: We could probably use the `index` value in conjunction with the hashed text (or hash `text + index`), since the passages can't be re-ordered. This would guarantee uniqueness of the key.

PR #417, opened 2022-09-26
fix: finish updating pdfjs

ColbyJohnIBM:

- Add prop to send in a url for pdf.js worker
- Use this prop in the example app and Storybook
- Adjust tests accordingly

Yes, any use of the `PdfViewer` component (standalone, or as part of `PdfViewerWithHighlight`, `DocumentPreview` or `PreviewDocument` will need to supply a URL string to a hosted `pdf.js` worker file (e.g. https://github.com/mozilla/pdfjs-dist/blob/master/build/pdf.worker.min.js) via the `pdfWorkerUrl` prop.

### DIFF
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
@@ -226,7 +226,8 @@ function _renderPage(
 let currentPdfWorkerUrl: string | null = null;
 function setupPdfjs(pdfWorkerUrl: string): void {
   // Only load the worker the first time the worker url is sent in
-  if (pdfWorkerUrl !== currentPdfWorkerUrl) {
+  // and don't load the worker in unit tests (see setupTests.ts)
+  if (typeof Worker !== 'undefined' && pdfWorkerUrl !== currentPdfWorkerUrl) {
     const pdfjsWorker = new Worker(pdfWorkerUrl);
     PdfjsLib.GlobalWorkerOptions.workerPort = pdfjsWorker;
     currentPdfWorkerUrl = pdfWorkerUrl;
--- a/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
+++ b/packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
@@ -57,7 +57,7 @@ export type PdfViewerProps = PdfDisplayProps & {
   /**
    * URL of hosted PDF worker
    */
-  pdfWorkerUrl?: string;
+  pdfWorkerUrl: string;
 };
 
 const PdfViewer: FC<PdfViewerProps> = ({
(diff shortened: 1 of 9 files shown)


packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -57,6 +54,10 @@ export type PdfViewerProps = PdfDisplayProps & {
    * Callback for text layer info
    */
   setRenderedText?: (info: PdfRenderedText | null) => any;
+  /**
+   * URL of hosted PDF worker
+   */
+  pdfWorkerUrl?: string;
```
noah-eigenfeld, line 60: If adding this url is a breaking change, then why is the prop optional? Will this component still work without the url? That goes for this component, but also any of the other components in this PR where we are passing through the url.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.stories.tsx
```diff
@@ -328,6 +332,8 @@ storiesOf('DocumentPreview/components/PdfViewerWithHighlight', module)
         highlights={highlightKnob.data[highlights]}
         activeIds={activeIds}
         setCurrentPage={handleSetCurrentPage}
+        // pulled from pdfjs-dist (see main.js > staticDirs)
+        pdfWorkerUrl={'pdf.worker.min.js'}
```
noah-eigenfeld, line 336: Nit: can we declare this url at the top of the file and pass it in as a var everywhere else?

noah-eigenfeld (commented):
One question I want answered before approving

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/__tests__/PdfViewer.test.tsx
```diff
@@ -7,6 +7,7 @@ import { document as doc } from 'components/DocumentPreview/__fixtures__/Art Eff
 
 describe('PdfViewer', () => {
   it('renders PDF document', async () => {
+    // @ts-ignore pdfWorkerUrl is required, but needs to be undefined for tests
```
noah-eigenfeld, line 10: This feels weird. If we have to twist our component in a way it shouldn't be used in order to test it, then that sounds like a bad test case. Why can't you pass in a blank or dummy string?

ColbyJohnIBM: The unit tests has always used the pdf.js `GlobalWorkerOptions.workerSrc` prop rather than the `GlobalWorkerOptions.workerPort` prop. This used to be defined in the production code. In this PR, I moved that out to the `setupTests` file so that all tests that use PdfViewer will have a worker set up already. Adding a `pdfWorkerUrl` prop to this component would set up the `GlobalWorkerOptions.workerPort` prop, which could break the worker (although I haven't actually tried it, so I'm not sure of that).

I can definitely add a more substantial comment here (and point to `setupTests`) if that would make the code clearer.

noah-eigenfeld (approved):
Expecting another commit with some additional changes that Colby and I discussed in DMs. With those, approved.

PR #467, opened 2023-01-21
fix: callback pdf render error

dorianmiller:

Problem when a PDF fails to render, the document view is blank. The fix defaults to display the text version.

Will add a unit test to load an invalid pdf file. 

No


examples/discovery-search-app/src/ExampleDocumentProvider.js
```diff
@@ -10,12 +10,31 @@
  */
 export class ExampleDocumentProvider {
   async provides(document: DocumentProviderProps) {
-    return document.extracted_metadata.filename === 'Art Effects.pdf';
+    // return document.extracted_metadata.filename === 'Art Effects.pdf';
+    console.log('dorian change', document.extracted_metadata.filename);
+    if (document.extracted_metadata.filename.indexOf('.pdf') >= 0) {
```
noah-eigenfeld, line 15: I don't know if this difference would ever matter, but you can also use `document.extracted_metadata.file_type === 'pdf'` as your check here (and wherever else you want to verify the document type). Honestly, since you've worked on the backend of Answers, you might have a better idea of how reliable the filename of a document that gets returned to us is.

examples/discovery-search-app/src/ExampleDocumentProvider.js
```diff
@@ -10,12 +10,31 @@
...
+      // hack to force PDF to render
+      document.extracted_metadata.text_mappings = true;
```
noah-eigenfeld, line 17: This seems like a bad practice. We shouldn't be modifying the object getting passed into this provider, because there might be other components or methods that assume this object is coming straight from the api response. If this started affecting the outcome of a completely different component in the future, it would be very hard to debug. We should try to fix how our code is reacting to bad data, or fix the example data at the example file we're feeding in.

Also, according to the typings we have in tooling, the `text_mappings` variable is either a string or an array of strings, but not a boolean.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
```diff
@@ -56,7 +56,11 @@ export type PdfViewerProps = PdfDisplayProps & {
    */
   setRenderedText?: (info: PdfRenderedText | null) => any;
   /**
-   * URL of hosted PDF worker
+   * Callback any errors on render
+   */
+  setIsPdfRenderError: (isError: boolean) => void;
+  /**
+   * Callback for text layer info
```
noah-eigenfeld, line 63: Why did you change the description of this variable? Shouldn't it still be `URL of hosted PDF worker`?

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
```diff
@@ -179,6 +190,29 @@ function useHighlightState({
...
+  const [currentIsPdfError, setCurrentIsPdfError] = useState(isPdfRenderError);
+  console.log('useIsPfdError', currentIsPdfError, callbackIsPdfError);
+
+  // Process error message
+  const _previouserrMsgFromPdf = useRef(isPdfRenderError);
+  useEffect(() => {
+    console.log('_previouserrMsgFromPdf', _previouserrMsgFromPdf);
+    console.log('currentErrMsgFromPdf  ', currentIsPdfError);
+    console.log('localErrMsgFromPdf    ', isPdfRenderError);
+    callbackIsPdfError?.(currentIsPdfError);
```
noah-eigenfeld, line 210: What is this line doing?

noah-eigenfeld (commented):
Some initial comments and questions

examples/discovery-search-app/src/ExampleDocumentProvider.js
```diff
@@ -10,12 +10,31 @@
...
+      console.log('dorian change, yes pdf');
```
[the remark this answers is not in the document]
dorianmiller: Thanks, I removed it.

examples/discovery-search-app/src/ExampleDocumentProvider.js
dorianmiller: I restored the original version. This was for local testing.

examples/discovery-search-app/src/ExampleDocumentProvider.js
dorianmiller: Good design point. Original file is restored.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewer.tsx
dorianmiller: Thanks for catching my copy-and-paste. I restored the original description.

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewerWithHighlight/PdfViewerWithHighlight.tsx
dorianmiller: Calling `callbackIsPdfError` is updating the state and passing the state change from PDFViewer (where error happened) to DocumentPreview (logic changes display from pdf to html/text). PdfViewerWithHighlight is an intermediate component.

noah-eigenfeld (approved):

PR #495, opened 2023-04-19
feat: update facet filtering logic

noah-eigenfeld:
Implementation updates for https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/15426

Updates the logic used when compiling a filter string from the selected facets, such that each sub-category of a facet is treated with AND logic instead of combining into ORs for the whole parent category.

1. Create a project with a set of documents and facets such that several facets will only apply to specific subsets
2. Test that within a subcategory, fields are filtered as ORs (`|` when viewing the network query call)
3. Test that between subcategories, fields are filtered as ANDs (`,` when viewing the network query call)
4. Test that fields between an uncategorized facet and a categorized facet still filter as ANDs
  ```
  Example
  A --- parent category
    B --- sub-category
      C --- term
      D --- term
    E --- sub-category
      F --- term
  G --- parent category
    H --- term


When C and D are selected, the documents returned are the union of C and D documents.
When C and F are selected, the documents returned are the intersection of C and F.
When C, D, F, and H are selected, the documents returned follow: (C OR D) AND F AND H
and so on...
  ```
N/A


packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
```diff
@@ -93,10 +93,33 @@ export class SearchFilterTransform {
     const filterStrings: string[] = [];
     facets.forEach(facet => {
       const field = get(facet, 'field', '');
-      const results = get(facet, 'results', []);
-      const keys = this.quoteSelectedFacets(results, 'key');
-      if (keys.length) {
-        filterStrings.push(keys.map(key => `${escapeFieldName(field)}::${key}`).join('|'));
+      const results: SelectableQueryTermAggregationResult[] | never[] = get(facet, 'results', []);
```
jhpedemonte, line 96: Why `never`?

jhpedemonte: Also, let's get rid of `get()` here:
```suggestion
      const results: SelectableQueryTermAggregationResult[] | never[] = facet?.results || [];
```

That also makes TS work better.

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
```diff
@@ -93,10 +93,33 @@ export class SearchFilterTransform {
...
+      const hasCategories = // logic duplicated from CollapsibleFacetsGroup
```
jhpedemonte, line 97: If the logic is duplicated, it makes sense to refactor this into a utility method that can be used in both places. Makes code more maintainable and avoids issue where not all of the "instances" of this code are updated at the same time, in the future.

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
```diff
@@ -93,10 +93,33 @@ export class SearchFilterTransform {
...
+        results?.[0]?.aggregations !== undefined;
+      // if a group of facets contain sub-categories, surround each sub-category with an AND (,)
+      if (hasCategories) {
+        const resultsByCategory: { [key: string]: SelectableQueryTermAggregationResult[] } = {};
+        results.forEach(result => {
+          const resultCategory = result!.aggregations![0].results![0].key;
+          if (resultCategory in resultsByCategory) {
+            resultsByCategory[`${resultCategory}`].push(result);
+          } else {
+            resultsByCategory[`${resultCategory}`] = [result];
```
jhpedemonte, line 109:
```suggestion
            resultsByCategory[resultCategory].push(result);
          } else {
            resultsByCategory[resultCategory] = [result];
```

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
```diff
@@ -93,10 +93,33 @@ export class SearchFilterTransform {
...
-      if (keys.length) {
-        filterStrings.push(keys.map(key => `${escapeFieldName(field)}::${key}`).join('|'));
+      const results: SelectableQueryTermAggregationResult[] | never[] = get(facet, 'results', []);
+      const hasCategories = // logic duplicated from CollapsibleFacetsGroup
+        field.includes('enriched_') &&
+        field.includes('entities.text') &&
+        results?.[0]?.aggregations !== undefined;
+      // if a group of facets contain sub-categories, surround each sub-category with an AND (,)
+      if (hasCategories) {
+        const resultsByCategory: { [key: string]: SelectableQueryTermAggregationResult[] } = {};
```
jhpedemonte, line 103: Nit: I've been finding that using `Record<>` makes the code a bit more readable:
```suggestion
        const resultsByCategory: { Record<string, SelectableQueryTermAggregationResult[]> } = {};
```

jhpedemonte (approved):
Code looks good. With suggestions fixed, approved.

packages/discovery-react-components/src/components/SearchFacets/utils/searchFilterTransform.tsx
noah-eigenfeld: 👍🏻 turns out I could get rid of the `never` typing without an error popping up

noah-eigenfeld: 👍🏻 turns out I could get rid of the `never` typing without an error popping up

PR #536, opened 2023-11-14
chore: upgrade `ibm-watson` and other deps

jhpedemonte:

- Upgrade `ibm-watson`, which allows us to upgrade `axios` to pick up vulnerability fixes.
    - BREAKING CHANGE: This upgrade causes types to change and may break consumers.
- Fixed issue where `ibm-watson` Disco query API was expecting prop `collectionIds` but we were passing `collection_ids`
- Upgrade to using Node v20
- Upgrade to yarn v4
- Use `vite` to build the components
- Fix storybook
- Upgrade a bunch of other dependencies

- Run storybook
- Run example app
- Run tooling with these changes

Yes, the `ibm-watson` upgrade changes some types, which may cause breaks in consuming code.


noah-eigenfeld (approved):
I stuck with storybook for testing, and didn't find any issues

PR #550, opened 2024-02-13
fix: fixed error in search results due to document title

aryashah27:
When using search, regardless of user query (empty or with actual strings), the message `There was an error rendering SearchResults`, when the code doesn't find document title where it is supposed to be. With this fix, it handles the case where title is its own separate field in the document with the format `"title": { "a": 1, "text": "title text" }`. If the field title is an object and doesn't have a field `text` in string format, just returns the `toString()` so that would result in displaying [object Object]. This would prevent the component from crashing.

Github issue: [16447](https://github.ibm.com/Watson-Discovery/disco-issue-tracker/issues/16447)

1. Download [test.json.zip](https://github.ibm.com/Watson-Discovery/disco-issue-tracker/files/1174563/test.json.zip)
2. Unzip the file
3. Create a collection by uploading test.json
4. Once the collection is created, go to `Improve and Customize`
5. No need to type something in the search bar, just hit enter
6. You should get the results with the `title text` as a part of the results

### DIFF
--- a/packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
+++ b/packages/discovery-react-components/src/components/SearchResults/components/Result/Result.tsx
@@ -129,10 +129,6 @@ export const Result: React.FunctionComponent<ResultProps> = ({
 
   let title = getDocumentTitle(result, resultTitleField);
 
-  if (Array.isArray(title)) {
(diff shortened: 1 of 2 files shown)


noah-eigenfeld (approved):
Code looks good, and the fix seems to be working when I test locally

PR #559, opened 2024-03-20
feat: initial changes to allow components to build with react 18

noah-eigenfeld:


examples/discovery-search-app/package.json
```diff
@@ -36,7 +36,7 @@
     "ibm-watson": "^8.0.0",
     "node-polyfill-webpack-plugin": "^3.0.0",
     "proper-url-join": "^2.1.1",
-    "react": "^17.0.2",
+    "react": "^18.0.2",
```
noah-eigenfeld, line 39: This should probably still be 17, since that's what WD Tooling is at. We can change them both if we ever migrate tooling to react 18.

noah-eigenfeld:
I was overthinking things with the overrides and resolutions. Closing in favor of this simplified PR: https://github.com/watson-developer-cloud/discovery-components/pull/560

PR #578, opened 2024-06-04
fix: resolve pdfjs-4 build errors

noah-eigenfeld:
Updates the PdfViewerTextLayer such that it builds without producing any errors

1. `yarn run build` in the components package runs successfully
2. When running storybook, stories that use the PDF viewer are rendered correctly, and highlighting still works


packages/discovery-react-components/src/components/DocumentPreview/components/PdfHighlight/PdfHighlight.tsx
```diff
@@ -88,6 +88,8 @@ const PdfHighlight: FC<Props> = ({
 
   const { textDivs } = pdfRenderedText || {};
 
+  console.log('textDivs', textDivs);
```
noah-eigenfeld, line 91: TODO: remove once finished verifying textDivs

packages/discovery-react-components/src/components/DocumentPreview/components/PdfViewer/PdfViewerTextLayer.tsx
```diff
@@ -67,23 +68,29 @@ const PdfViewerTextLayer: FC<PdfViewerTextLayerProps> = ({
...
 
-          await _renderTextLayer(builder, textContent, textLayerDiv, scale);
-          return { textContent, viewport, page, textDivs: builder.textDivs };
+          await _renderTextLayer(builder, textLayerWrapper, viewport);
+          //TODO: is textLayerWrapper the right value for the container?
+          const pdfJsTextLayer = new PdfJsTextLayer({
+            textContentSource: textContent,
+            container: textLayerWrapper,
+            viewport
+          });
```
noah-eigenfeld, line 87: Main question to answer before this is mergable. TextLayer needs a `container`, and I am providing it with `textLayerWrapper`, which looks similar to what it's expecting. However, I'm seeing slightly different output in the `textDivs` styling this produces, so I'm checking whether that actually breaks anything and if we'll need to find a better `container`. Could also be a difference in the order in which I'm running renders that involve `textLayerWrapper`, since it gets instantiated inside `_renderTextLayer()` before I run the render on `PdfJsTextLayer`, which consumes it.

noah-eigenfeld:
I'm expecting the builds on this PR to fail until @dorianmiller's test updates are able to be merged into Javier's branch, and then this one is updated

packages/discovery-react-components/src/components/DocumentPreview/__tests__/DocumentPreview.test.tsx
```diff
@@ -251,8 +251,8 @@ describe('DocumentPreview', () => {
           </DiscoverySearch>
         );
 
-        // 58 pages evidence PDF rendered
-        const elem = await screen.findByText('58 pages', {
+        // evidence PDF rendered
+        const elem = await screen.findByText('“Acceptance Tests” or “Acceptance Testing”', {
```
noah-eigenfeld, line 255: This test was looking for an element of the toolbar UI, and not the PDF itself. Not sure why it started failing, but this is more faithful to the intention of the test, and it made it pass.

jhpedemonte (approved):

